PR #1, opened 2022-02-21
Create TRTools:v4.0.2

hopedisastro:
TRTools package is used to analyse and compare output across different short read STR callers.


vladsaveliev (approved):
Looks good to me!

PR #8, opened 2022-03-14
Add peddy DockerFile

vivbak:
Peddy required for [create_test_subset.py](https://github.com/populationgenomics/sample-metadata/pull/165).

### DIFF
--- a/images/peddy/Dockerfile
+++ b/images/peddy/Dockerfile
@@ -14,3 +14,4 @@ RUN apt-get update && apt-get install -y --no-install-recommends g++ libopenblas
         cpg-utils \
         peddy \
         sample-metadata 
+        
--- a/images/peddy/Dockerfile
+++ b/images/peddy/Dockerfile
@@ -3,7 +3,7 @@ FROM australia-southeast1-docker.pkg.dev/analysis-runner/images/driver-base:1.0
 
 ENV HAIL_QUERY_BACKEND service
 
-# hadolint ignore=DL3013
+# hadolint ignore=DL3008, DL3013
 RUN apt-get update && apt-get install -y g++ libopenblas-base liblapack3 openjdk-11-jre-headless python3-pip && \
     rm -r /var/lib/apt/lists/* && \
     rm -r /var/cache/apt/* && \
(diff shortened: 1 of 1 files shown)


images/peddy/Dockerfile
```diff
@@ -0,0 +1,16 @@
+# This image adds cpg-hail, SM and peddy to the base image.
+FROM australia-southeast1-docker.pkg.dev/analysis-runner/images/driver-base:1.0
+
+ENV HAIL_QUERY_BACKEND service
+
+# hadolint ignore=DL3008, DL3013
+RUN apt-get update && apt-get install -y --no-install-recommends g++ libopenblas-base liblapack3 openjdk-11-jre-headless python3-pip && \
```
illusional, line 7: I think python-3 and peddy versions should be pinned, the rest I'm happy to ignore

illusional (commented):
Any reason in particular it's added to the base-image, and not just debian - are you intending to use this as a driver image?

images/peddy/Dockerfile
```diff
@@ -0,0 +1,16 @@
...
+    rm -r /var/lib/apt/lists/* && \
+    rm -r /var/cache/apt/* && \
+    pip3 install --no-cache-dir \
+        analysis-runner \
+        bokeh \
+        cpg-hail \
```
vladsaveliev, line 13: Does it need hail?

vivbak:
@illusional, yeah I wanted to specify this image with `--image` for the analysis-runner. Although, I may have misunderstood the best way to run this file via the AR with the peddy dependency met?

images/peddy/Dockerfile
vivbak: Will do!

images/peddy/Dockerfile
vivbak: Good point, will scrap.

illusional:
This is probably an annoying request, but could we refactor peddy out of the [create test subset script](https://github.com/populationgenomics/sample-metadata/pull/165), it looks like it's only used to get family names - and that's the only addition that blocks that script from being used by the regular driver image - and I don't know if this Dockerfile contains google-cloud libraries either.

illusional:
Further, could these lines just be replaced by a `get_families` call?
https://github.com/populationgenomics/sample-metadata/blob/852b94fe2c3c2343bd1d813e879631cc6ac28f84/scripts/create_test_subset.py#L107-L110

vivbak:
Will refactor and open a new pr for the changes to create_test_subset

PR #9, opened 2022-05-09
REViewer Dockerfile

hopedisastro:


images/reviewer/Dockerfile
```diff
@@ -0,0 +1,16 @@
...
+wget --progress=dot:giga https://github.com/Illumina/REViewer/releases/download/v0.2.7/REViewer-v0.2.7-linux_x86_64.gz && \
+gunzip REViewer-v0.2.7-linux_x86_64.gz && \
+chmod +x REViewer-v0.2.7-linux_x86_64 && \
+apt-get clean && \
+rm -rf /var/lib/apt/lists/*
+
+# Clean up
+RUN rm -rf /tmp/* \
+    /var/tmp/* \
+    /var/cache/apt/*
```
illusional, line 15: Can you just join these runs together to save a layer - and I've added the gzipped file.

```suggestion
rm -rf /var/lib/apt/lists/* \
    /tmp/* \
    /var/tmp/* \
    /var/cache/apt/* \
    REViewer-v0.2.7-linux_x86_64.gz
```

illusional (changes requested):

illusional:
Hey @hopedisastro, could you also remove the .DS_Store files - and there is an additional linting issue. Let me know if you have any questions :)

hopedisastro (commented):
Changes made. Thank you @illusional :)

illusional (approved):
Looks good to me! I'll deploy this now :)

images/reviewer/Dockerfile
```diff
@@ -0,0 +1,14 @@
+FROM ubuntu:20.04
```
lgruen, line 1: @illusional Should this derive from the [existing driver base image](https://github.com/populationgenomics/analysis-runner/blob/main/driver/Dockerfile.base) or is there a better way to install the required dependencies?

lgruen (changes requested):

images/reviewer/Dockerfile
illusional: Only if it's going to be used as a driver image in the analysis runner - @hopedisastro, are you able to clarify if you're just using this as a job within a larger workflow, or planning to use the `--image` parameter in the analysis-runner.

hopedisastro: I think I will be using this as a job in Hail batch, to visualise read pileups of certain loci.

PR #13, opened 2022-05-10
str-r image Dockerfile

hopedisastro:
Image containing R packages for STR analysis.

### DIFF
--- a/images/str-r/Dockerfile
+++ b/images/str-r/Dockerfile
@@ -1,4 +1,4 @@
-FROM australia-southeast1-docker.pkg.dev/analysis-runner/images/driver-base:1.1
+FROM australia-southeast1-docker.pkg.dev/analysis-runner/images/driver-r:1.1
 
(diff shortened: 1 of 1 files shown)


images/str-r/Dockerfile
```diff
@@ -0,0 +1,23 @@
+FROM australia-southeast1-docker.pkg.dev/analysis-runner/images/driver-base:1.1
```
illusional, line 1: This should start from the `r` driver base:

```suggestion
FROM australia-southeast1-docker.pkg.dev/analysis-runner/images/driver-r:1.1
```
```suggestion
FROM australia-southeast1-docker.pkg.dev/analysis-runner/images/driver-base:1.1
```

illusional (changes requested):
Just need to base it off the `r-driver` image now - as the regular driver-base doesn't have any other packages.

illusional (approved):

PR #18, opened 2022-05-17
Fixes

vivbak:
* Update pip version to pip3.10
* Fix interactive install for python3.10
* Update python3-dev to python3.10 dev.

### DIFF
--- a/images/cpg-pipes/Dockerfile
+++ b/images/cpg-pipes/Dockerfile
@@ -3,7 +3,7 @@ FROM australia-southeast1-docker.pkg.dev/analysis-runner/images/driver-base:1.1
 
 ENV TZ=Australia \
     DEBIAN_FRONTEND=noninteractive
-# hadolint ignore=DL3008, DL3013
+# hadolint ignore=DL3008, DL3013, DL4006
 RUN apt-get update && apt-get install -y --no-install-recommends \
     software-properties-common \
     gcc \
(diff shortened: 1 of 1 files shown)


vladsaveliev:
Can we actually update `driver-base` to use python 3.10?

images/cpg-pipes/Dockerfile
```diff
@@ -1,13 +1,18 @@
...
     build-essential \
-    python3 \
-    python3-pip \
-    python3-dev \
+    curl \
     && \
-    pip install --no-cache-dir cpg-pipes==0.3.1 && \
+    add-apt-repository ppa:deadsnakes/ppa -y && \
+    apt-get install -y --no-install-recommends python3.10 python3.10-distutils python3-apt python3.10-dev  && \
+    curl -sS https://bootstrap.pypa.io/get-pip.py | python3.10 && \
```
vladsaveliev, line 15: Can't we install pip for python 3.10 with apt-get?

vladsaveliev: E.g. even if `python3.10` is installed, `python3-pip` would still somehow use python 3.8?

images/cpg-pipes/Dockerfile
```diff
@@ -1,13 +1,18 @@
...
+    pip3.10 install --no-cache-dir cpg-pipes==0.3.1 && \
```
vladsaveliev, line 16: Wondering if it's possible to make `pip` and `python` point to `pip3.10` and `python3.10` by default, so we don't need to worry about python versions when running python scripts

vladsaveliev:
Another option would be to change the base to `FROM python:3.10`

vladsaveliev (approved):

images/cpg-pipes/Dockerfile
vivbak: I had a really difficult time doing this, but very open to suggestions.

PR #25, opened 2022-06-14
Add validation tools

MattWellie:
Adds a skopeo script to copy the latest version of the hap.py docker image into the CPG registry

Adds a Dockerfile to build the RTGtools (`vcfeval`) into that same image, using the standalone jar file from https://www.realtimegenomics.com/products/rtg-tools

### DIFF
--- a/images/validation/Dockerfile
+++ b/images/validation/Dockerfile
@@ -1,0 +1,16 @@
+FROM australia-southeast1-docker.pkg.dev/cpg-common/images/hap.py:v0.3.9
+
+ENV TZ=Australia
+
+RUN apt-get update && \
+    apt-get install -y unzip && \
+    apt-get clean -y
+
+
+# Install VcfEval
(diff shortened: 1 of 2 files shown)


images/validation/Dockerfile
```diff
@@ -0,0 +1,14 @@
+FROM pkrusche/hap.py:v0.3.9
```
illusional, line 1: Wouldn't do from our internal one that we're going to copy?

illusional (changes requested):
Could you give a more descriptive name than validation, could be happy-rtg?
\+ Linting

images/validation/Dockerfile
MattWellie: nice catch, I'd originally had that before swapping back to test the build locally

illusional (approved):

vladsaveliev (approved):

PR #29, opened 2022-08-08
Happy 0.3.15

vladsavelyev:
Use the Dockefile hosted on the Illumina hap.py repo to build the image, update hap.py to 0.3.15


vladsavelyev:
The image contains RTG and bcftools:

/opt/hap.py/libexec/rtg-tools-install/RTG.jar
/opt/hap.py/bin/bcftools
/opt/hap.py/bin/hap.py
/opt/hap.py/bin/pre.py

Though it's annoying that PATH is not set to `/opt/hap.py/bin`. I'll add a comment to the README rather than have a modified version of Dockerfile.

MattWellie:
How would this be used to generate the image in the CPG repo?

vladsavelyev:
@MattWellie, actually manually using the commands in README. There is no Dockerfile here, so the github can't be used. We could create another CI job that runs a bash script instead of building a Dockerfile, not sure 🤔

PR #32, opened 2022-08-08
no braces, slight tweak to arg passing

MattWellie:
The variable expansion for `${VERSION}` inside the docker build is still coming out as `***VERSION***` instead of 0.3.15, even though the Docker instructions suggest this should be passed correctly:

see https://github.com/populationgenomics/images/runs/7721337852?check_suite_focus=true

https://docs.docker.com/engine/reference/builder/#understand-how-arg-and-from-interact

```
docker build --tag $DOCKER_IMAGE images/$IMAGE_NAME/ --build-arg VERSION=$IMAGE_TAG
  shell: /usr/bin/bash -e ***0***
  env:
    DOCKER_BUILDKIT: 1
    BUILDKIT_PROGRESS: plain
    CLOUDSDK_CORE_DISABLE_PROMPTS: 1
    DOCKER_IMAGE: australia-southeast1-docker.pkg.dev/cpg-common/images/hap.py
    IMAGE_NAME: hap.py
    IMAGE_TAG: 0.3.15
```

I think there's something I'm not seeing here, so I'd really welcome eyes/improvements!

### DIFF
--- a/images/hap.py/Dockerfile
+++ b/images/hap.py/Dockerfile
@@ -1,4 +1,3 @@
-ARG VERSION=${VERSION:-v0.3.15}
 FROM ubuntu:18.04
 
 # installation from https://github.com/Illumina/hap.py/blob/master/Dockerfile
@@ -43,7 +42,7 @@ RUN wget http://archive.apache.org/dist/ant/binaries/apache-ant-1.9.7-bin.tar.gz
 ENV PATH="$PATH:/opt/apache-ant-1.9.7/bin"
 
 # clone in, install, delete
-RUN git clone --depth 1 --branch $VERSION https://github.com/Illumina/hap.py /opt/hap.py-source && \
+RUN git clone --depth 1 --branch v0.3.15 https://github.com/Illumina/hap.py /opt/hap.py-source && \
 	python /opt/hap.py-source/install.py /opt/hap.py --with-rtgtools --no-tests && \
 	rm -rf /opt/hap.py-source
 
--- a/images/hap.py/Dockerfile
+++ b/images/hap.py/Dockerfile
@@ -1,5 +1,7 @@
 FROM ubuntu:18.04
 
+ARG VERSION=${VERSION:-v0.3.15}
+
 # installation from https://github.com/Illumina/hap.py/blob/master/Dockerfile
 ENV DEBIAN_FRONTEND=noninteractive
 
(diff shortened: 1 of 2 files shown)


lgruen:
Are you sure this isn't just GitHub Actions redacting the concrete value as a security hardening measure (to avoid logging secrets like passwords accidentally)?

MattWellie:
I thought it could be something like that, but in the Docker build Environment it publishes the values in plain text. If the values can be set as Env variables, I'm surprised the GitHub action would be able to redact them afterwards

MattWellie:
Likely explanation found by @lgruen: 

```
An ARG declared before a FROM is outside of a build stage, so it can’t be used in any instruction after a FROM. To use the default value of an ARG declared before the first FROM use an ARG instruction without a value inside of a build stage:
```

I was grabbing the version using ARG prior to the FROM argument (even though it was not used in the FROM line). After the FROM all set ARG variables are purged from the build environment. Pulling the image tag after the FROM should make it available during the rest of the build.

Every day is a school day

images/hap.py/Dockerfile
```diff
@@ -40,10 +41,10 @@ WORKDIR /opt
 RUN wget http://archive.apache.org/dist/ant/binaries/apache-ant-1.9.7-bin.tar.gz && \
     tar xzf apache-ant-1.9.7-bin.tar.gz && \
     rm apache-ant-1.9.7-bin.tar.gz
-ENV PATH $PATH:/opt/apache-ant-1.9.7/bin
+ENV PATH="$PATH:/opt/apache-ant-1.9.7/bin"
 
 # clone in, install, delete
-RUN git clone --depth 1 --branch v${VERSION} https://github.com/Illumina/hap.py /opt/hap.py-source && \
+RUN git clone --depth 1 --branch ${VERSION} https://github.com/Illumina/hap.py /opt/hap.py-source && \
```
lgruen, line 47:
```suggestion
RUN git clone --depth 1 --single-branch --branch ${VERSION} https://github.com/Illumina/hap.py /opt/hap.py-source && \
```

lgruen (approved):

images/hap.py/Dockerfile
```diff
@@ -1,6 +1,7 @@
-ARG VERSION=0.3.15
 FROM ubuntu:18.04
```
lgruen, line 1:
```suggestion
FROM ubuntu:22.04
```

PR #33, opened 2022-08-09
Revert to earlier ubuntu for python2

MattWellie:
Finally managed to build this locally! I've tested that `hap.py` and `pre.py` are both in $PATH, and the env variable $RTG_JAR is set and functional

Confirmed with the Hap.py codebase that the python version supported is <=2.7.3, so we need an install environment for python2. As a result I've reverted the base image Ubuntu 22.04 -> 18.04.

It did get stuck on the Hap.py/RTG install during my first local attempt, then on a retry it built smoothly (one to watch on the deploy pipeline)

### DIFF
--- a/images/hap.py/Dockerfile
+++ b/images/hap.py/Dockerfile
@@ -1,4 +1,4 @@
-FROM ubuntu:22.04
+FROM ubuntu:18.04
 
 ARG VERSION=${VERSION:-v0.3.15}
 


lgruen (approved):
> Confirmed with the Hap.py codebase that the python version supported is <=2.7.3, so we need an install environment for python2.

Wow... Is it still officially maintained? If not, I wonder if we should be using it at all?

MattWellie:
> Wow... Is it still officially maintained? If not, I wonder if we should be using it at all?

I confirmed that completely in error. It was too early a start for me today. The installation checks that 2.7.3 is a _minimum_ (sorry), though the code is Python2 only, and the issues page does have a huge number of requests for Python3 going back to 2018.

I don't think we can consider Hap.py to be in active development, but during the installation process the comparison engine (RTG tools) is downloaded, and the version there is up to date.

https://github.com/Illumina/hap.py/blob/master/external/make_dependencies.sh#L112-L143

illusional:
Hey @MattWellie, there are linting errors on this PR - could you fix in a follow-up PR please 🙏

PR #37, opened 2022-08-16
Update sam and bcftools

MattWellie:
The latest change of the hap.py image was to bring the hap.py codebase up to date (inc. the bundled version of RTG tools)

This has not updated the SAM/BCFtools versions which are included, currently 1.4.1 (2017)

This tweak to the build file updates the versions of both tools to 1.15.1, and adds a configurable build hts_version argument to update again in future

Tested on a local build - bcftools and samtools are both distributed with an internal HTSlib that they are compiled with, so this should not have any downstream implications

There is at least one command for Hap.py that uses a deprecated argument, but BCFtools transparently changes this to the correct form and issues a warning

### DIFF
--- a/images/hap.py/Dockerfile
+++ b/images/hap.py/Dockerfile
@@ -1,7 +1,7 @@
 FROM ubuntu:18.04
 
 ARG VERSION=${VERSION:-v0.3.15}
-ARG HTS_VERSION=1.15.1
+ARG HTS_VERSION==${HTSVERSION:-1.15.1}
 
 # installation from https://github.com/Illumina/hap.py/blob/master/Dockerfile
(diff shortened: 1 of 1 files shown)


cassimons (commented):
LGTM but I will leave approval to Vlad.

vladsavelyev:
I'm okay to trust it works, but personally at this point I'd switch to a bioconda distribution 🤔

vladsavelyev (approved):

MattWellie:
@vladsavelyev  I think I agree... the way that hap.py has bundled all the tools into git, then bot updated the binaries for 5 years is surprising and disappointing...

PR #40, opened 2022-08-26
Add vep image

vladsavelyev:


vladsavelyev:
Hmm, do we need the linter to be that strict? I might be naive, but none of the inspections that triggered warnings look particularly useful

lgruen:
> Hmm, do we need the linter to be that strict? I might be naive, but none of the inspections that triggered warnings look particularly useful

Yeah, if there's an override for the default warnings, that would be great.

lgruen (approved):

PR #44, opened 2022-10-19
updates the loftee plugin

MattWellie:
The `main` branch of loftee causes a ton of error logging messages when running on GRCh38.
Example [here](https://batch.hail.populationgenomics.org.au/batches/378307/jobs/17)  (similar errors being throw 27000+ times during processing of one variant subset)

There is a separate branch which we can use instead, with an updated version of the plugin files, see https://github.com/konradjk/loftee/issues/84 (and other similar issues)

The path to copy the plugin files to contains a `*` - 

 -  for VEP 105.0, the path is /root/micromamba/share/ensembl-vep-105.0-1
 - for 107.0 this is /root/micromamba/share/ensembl-vep-107.0-0

This copy works for both situations when building the image, but doesn't resolve hard-coding like [this](https://github.com/populationgenomics/production-pipelines/blob/main/jobs/vep.py#L194)

As of today I don't believe there's a way to set an env variable to the result of a RUN command, otherwise this could be used:

`ls -l ${MAMBA_ROOT_PREFIX}/share/ | grep ensembl | awk '{ print $9}'`

this could be used in setting the path in the production-pipelines job?

```
LOFTEE_PLUGIN_PATH="${MAMBA_ROOT_PREFIX}/share/$(ls -l ${MAMBA_ROOT_PREFIX}/share/ | grep ensembl | awk '{ print $9 }')"
```

... but this looks really heavy handed

### DIFF
--- a/images/vep/Dockerfile
+++ b/images/vep/Dockerfile
@@ -2,7 +2,7 @@ FROM debian:stretch-slim
 
 ENV MAMBA_ROOT_PREFIX=/root/micromamba
 ENV PATH=$MAMBA_ROOT_PREFIX/bin:$PATH
-ARG VERSION=${VERSION:-104.3}
+ARG VERSION=${VERSION:-105.0}
 
 # hadolint ignore=DL3008
 RUN apt-get update && \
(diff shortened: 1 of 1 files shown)


images/vep/Dockerfile
```diff
@@ -2,27 +2,43 @@ FROM debian:stretch-slim
...
-    apt-get install -y git wget bash bzip2 zip lsb-release curl gnupg2 && \
+    apt-get install --no-install-recommends -y \
+    bash \
+    bzip2 \
+    curl \
+    git \
+    gnupg2 \
+    lsb-release \
+    wget \
+    zip \
```
vladsavelyev, line 17: Can't see if it's just reformatted/reodered or a package added here?

vladsavelyev (approved):

images/vep/Dockerfile
MattWellie: ah yeah, this was just splitting the packages out and ordering alphabetically - no changes

MattWellie:
I'm going to merge/build this, and if it works well I'll look into the CI fixes (also broken on main, and I'm reluctant to change the working build)

PR #46, opened 2022-10-19
Fixes to LOFTEE plugin in VEP build

MattWellie:
Quite a few alterations here... Please review

1. LoF doesn't compile due to `pm`/`pl` files being unable to find each other, see https://github.com/konradjk/loftee/issues/60. Solution used - https://github.com/populationgenomics/loftee_38, a fork of grch38 from LOFTEE which moves all required files into the same folder, and updates import paths accordingly. Remove all loftee plugin content from the default VEP content, and added all files manually from the github clone. Adds this path as env PERL5LIB (see https://github.com/konradjk/loftee/issues/84#issuecomment-1025776382)

2. Perl module BigFile for `.bw` queries is not installed - added `perl-bio-bigfile` to the conda installation (see https://github.com/sigven/cpsr/issues/21

3. LoF doesn't compile due to BioPerl being unavailable
```Failed to compile plugin LoF: Can't locate Bio/Perl.pm in @INC ```
The `RUN vep_install --AUTO a --NO_HTSLIB` line of the build should solve this. The `vep_install` script runs a check for a later VEP version unless it's explicitly told not to. Now that VEP 108 has been released, this breaks the non-interactive install command. Adding `--NO-UPDATE` fixes this

Locally on a small VCF (50 variants) I've been able to run this in remote DB mode, using a correctly compiled LoF plugin to generate some LOFTEE results in the output.

cmd used: 
```
vep --database -i /data/50_variants.vcf.gz -o /data/out.vcf --assembly GRCh38 --dir_plugins /root/micromamba/share/ensembl-vep-105.0-1 --plugin LoF,loftee_path:/root/micromamba/share/ensembl-vep-105.0-1 --force_overwrite --no_stats
```

### DIFF
--- a/images/vep/Dockerfile
+++ b/images/vep/Dockerfile
@@ -58,5 +58,5 @@ ENV PERL5LIB=$MAMBA_ROOT_PREFIX/share/ensembl-vep-${VERSION}-1
 
 # Install Bio::Perl for Loftee (conda's perl-bioperl doesn't work)
 # run this last, as other commands seem to re-break the installation
-RUN vep_install --AUTO a --NO_HTSLIB
+RUN vep_install --AUTO a --NO_UPDATE --NO_HTSLIB
 
--- a/images/vep/Dockerfile
+++ b/images/vep/Dockerfile
@@ -3,6 +3,7 @@ FROM debian:stretch-slim
 ENV MAMBA_ROOT_PREFIX=/root/micromamba
 ENV PATH=$MAMBA_ROOT_PREFIX/bin:$PATH
 ARG VERSION=${VERSION:-105.0}
+ENV VEP_SHARE=$MAMBA_ROOT_PREFIX/share/ensembl-vep-${VERSION}-1
 
 # hadolint ignore=DL3008
 RUN apt-get update && \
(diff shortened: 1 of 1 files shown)


lgruen (approved):
I'm really not a VEP expert, but this looks fine to me :)

vladsavelyev:
Sorry I'm late to review. Thanks for working on this! Do we need to change the `vep` command that we use with Hail Batch and dataproc scripts (e.g. in [seqr-loader](https://github.com/populationgenomics/production-pipelines/blob/6ddc40906deb0dcb6e30386a458e61df6c57e435/jobs/vep.py#L197) and [dataproc config](https://github.com/populationgenomics/production-pipelines/blob/from-initial/scripts/vep/dataproc-config.json)?

MattWellie:
@vladsavelyev I think it should be a clean change, as the image is still being built on 105.0 so lines like [this](https://github.com/populationgenomics/production-pipelines/blob/6ddc40906deb0dcb6e30386a458e61df6c57e435/jobs/vep.py#L194) will be fine. 

It does raise the question of whether we should update to 107/108. If/when we do update, there might be a few changes required across the pipeline codebase(s)

Also I've noticed a discrepancy in those two files you linked to, dataproc has `/usr/local/share/ensembl-vep-__VEP_VERSION__-0`, the other has `LOFTEE_PLUGIN_PATH=/root/micromamba/share/ensembl-vep-105.0-1`, which to my eye means the dataproc path is not correct

PR #50, opened 2022-11-02
Add `images.toml` and CI to rebuild images from the config

vladsavelyev:
* Add `images.toml` listing available images
* Add CI to automatically build images for tags in TOML


illusional (approved):
Oh, so this writes a set of outputs, and the deploy job is ran for reach of the outputs.

Are all of the images defined in the TOML present in this repo?
Otherwise just some documentation on the README for this change would be awesome :)

lgruen (approved):
This is a nice idea! Related to Michael's comment: should we enforce that the TOML lists *exactly* the same images that are available in this repo and use it as the main mechanism to deploy all the images? You'd just need to add a `glob` to make sure the entries in `images.toml` cover exactly all the available images.

illusional:
I think it could support both, images NOT listed in the toml, but also trigger rebuild for images that don't necessarily change the tags (we do this sometimes for multiply when the driver changes).

vladsavelyev:
Oh, haven't noticed that you approved! ❤️ 

I added into the `images.toml` all other images from this repo. We should probably add all of the other ones from `cpg-commons/images` that were not present in this repo, but there are so many, perhaps we should review if we want to remove some of those. There are also gatk-sv images prefixed with `sv/` which I handle separately, but I'll add them in a separate PR in the future.

I added some docs as suggested.

What do you think about the images that we move with `skopeo`? Since there are no Dockerfiles, but perhaps we should support adding images with shell scripts using the same CI? Though looked at those, they are all available on bioconda, and we should be able to use the same pattern with miniconda and add proper Dockefiles. Done than now, and also updated the docs to recommend restrain from using `skopeo` if possible.

vladsavelyev:
There is a bunch of images that are not single-software, as the purpose of this repo is stated in the README:

```
This is a repository of:

* Single-software images to build and push in the `cpg-common` artifact registry.

This is a BAD place for:

* multi-software images
```

Namely:

* `str-r` is multi-tool and cannot be properly tagged following a specific tool version (added by Home)
* `hail-gcp-notebook` is multi-tool, though theoretically it can be versioned by hail version. However, the build with the current Dockerfile [fails](https://github.com/populationgenomics/images/actions/runs/3435927375/jobs/5728868187) @lgruen 
* `stripy` is a multi-tool pipeline. It is versioned by the author using git tags, however the latest tag there is v2.1, and it doesn't match the Dockefile's v2.2 @MattWellie.
 
So we won't be able to add all images into `images.toml` and support by automated CI. What should we do:
* change the purpose of this repo and keep some images for manual re-deploy, 
* or move those images elsewhere?

MattWellie:
> stripy is a multi-tool pipeline
Yes, but I don't think it's that simple. The tool being used is Stripy, and everything else is just a prerequisite to run it. All those other tools are pinned to a specific version, so the only variable is STRipy. All of these images are multi-tool, but most of the tools are installed using apt-get, so we don't consider them 'multi-tool'.

> the latest tag there is v2.1, and it doesn't match the Dockefile's v2.2
that's really weird, their own build files/tags all point to v2.2... which they haven't published
[latest changes](https://gitlab.com/andreassh/stripy-pipeline/-/commit/a7e65238db0acf20c8b159663c7bf05d15c24bf2)

vladsavelyev:
> Yes, but I don't think it's that simple. The tool being used is Stripy, and everything else is just a prerequisite to run it. All those other tools are pinned to a specific version, so the only variable is STRipy. All of these images are multi-tool, but most of the tools are installed using apt-get, so we don't consider them 'multi-tool'.

Well, I certainly wouldn't talk about the software installed with apt-get as tools in this context. What I meant, is that Stripy looks more like a driver for ExpansionHunter and REViewer. But we certainly can look at it as a single-tool Stripy image, but the issue is that the v2.2 tag is missing

PR #52, opened 2022-11-07
add STRipy Dockerfile

MattWellie:
I spent a while this morning building up a dockerfile, then realised that STRipy already has a dockerfile... just not in the STRipy repo 🙃 

sourced from https://stripy.org/Dockerfile, with a samtools version bump

### DIFF
--- a/images/stripy/Dockerfile
+++ b/images/stripy/Dockerfile
@@ -1,0 +1,69 @@
+FROM ubuntu:20.04
+MAINTAINER Alena Musilova, alena.musilova@lfmotol.cuni.cz
+
+LABEL \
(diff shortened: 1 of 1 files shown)


cassimons (approved):
Great thanks.  Is this all that is needed for the pipeline version as well?

MattWellie:
@cassimons This dockerfile was taken from this pointer:
https://gitlab.com/andreassh/stripy-pipeline/-/blob/main/runDocker.sh#L49-52

which should set up the container to start inside the `stripy-pipeline` folder, so you can either run the container interactively, or just do a `docker run python3 stri.py ...`

PR #54, opened 2022-11-17
add gcloud to stripy

cassimons:
Add the gcloud sdk to the stripy image. For right or wrong, I have done this install as per google's instructions not the conda aproach cpg seems to typically use.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -22,10 +22,10 @@ repos:
       - id: black
 
   - repo: https://github.com/PyCQA/flake8
-      rev: '5.0.4'    
-      hooks:
-        - id: flake8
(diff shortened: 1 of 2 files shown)


MattWellie:
The main reason afaik for our other install process seems to be the debian package being out of date, but I don't know if that is still true...

https://github.com/populationgenomics/analysis-runner/blob/main/driver/Dockerfile.base#L19

That driver image is also based off Ubuntu 22, and this is 20. Not specific to this change but that could be worth updating?

cassimons:
Fixed both of these. Linting is failing due to flake repo change but have PR in for that.

MattWellie (approved):
gcloud FTW

PR #56, opened 2022-11-18
CI to copy config to `gs://cpg-config/templates`

vladsavelyev:


illusional (approved):
Do we need to give the image deployer write access to that bucket, or will it already have it.

Also noticed the CI is broken for this repo too

.github/workflows/toml-changed.yaml
```diff
@@ -80,3 +80,7 @@ jobs:
       - name: "push image"
         run: |
           docker push "${PREFIX}/${{ matrix.name }}:${{ matrix.tag }}"
+
+      - name: "deploy config"
+        run: |
+          gsutil cp images.toml gs://cpg-config/templates/images/images.toml
```
lgruen, line 86:
```suggestion
          gcloud storage cp images.toml gs://cpg-config/templates/images/images.toml
```

lgruen (approved):

vladsavelyev:
> Also noticed the CI is broken for this repo too

CI checks the differences in images.toml between the previous revision and the current revision. That's why I split my previous changes into 2 stacked PRs: one adds images.toml, another one adds CI that checks the differences. You accidentally merged the second one into the first one, so after we merged them altogether, it broke the first CI run :) But consequent merges should be alright.

vladsavelyev:
> Do we need to give the image deployer write access to that bucket, or will it already have it.

Yeah I think we need to do that, good spot

vladsavelyev:
We probably also want to read infra-specific parameters like `PREFIX: gs://cpg-config/templates`, `PREFIX: australia-southeast1-docker.pkg.dev/cpg-common/images` from somewhere, e.g. secrets deployed in Pulumi. What do you think @illusional @lgruen

lgruen:
I think that would be handy, as it could be used both by this CI (to write blobs) and the analysis-runner (to read blobs).

vladsavelyev:
@lgruen @illusional could you please point me to where the service account used to deploy images here is created?

PR #63, opened 2022-12-05
install latest CellRegMap from github in image

annacuomo:
I need to use CellRegMap functions that are not yet released to PyPI, so installing directly from github (but not 100% sure of the syntax!). Thanks!

### DIFF
--- a/images/cellregmap/Dockerfile
+++ b/images/cellregmap/Dockerfile
@@ -1,10 +1,11 @@
-ARG DRIVER_TAG="latest"
+# ARG DRIVER_TAG="latest"
+ARG DRIVER_TAG="dev"
 
 FROM australia-southeast1-docker.pkg.dev/analysis-runner/images/driver:$DRIVER_TAG
 
 SHELL ["/bin/bash", "-o", "pipefail", "-c"]
 
-ARG VERSION=${VERSION:-0.0.3}
+# ARG VERSION=${VERSION:-0.0.3}
 
 # hadolint ignore=DL3008,DL3013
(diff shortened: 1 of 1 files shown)


images/cellregmap/Dockerfile
```diff
@@ -14,7 +15,7 @@ RUN apt-get update && \
     # The chi2comb-py Python package depends on the chi2comb C++ library.
     curl -fsSL https://raw.githubusercontent.com/limix/chi2comb/master/install | bash && \
     pip3 install --no-cache-dir \
-        cellregmap==${VERSION} \
+        cellregmap==git+https://github.com/annacuomo/CellRegMap \
```
illusional, line 18:
```suggestion
        git+https://github.com/annacuomo/CellRegMap \
```

images/cellregmap/Dockerfile
```diff
@@ -1,10 +1,11 @@
-ARG DRIVER_TAG="latest"
+# ARG DRIVER_TAG="latest"
+ARG DRIVER_TAG="dev"
```
illusional, line 2: This is the analysis-runner DRIVER_TAG, uses below, don't change this.

illusional (changes requested):

images/cellregmap/Dockerfile
annacuomo: ok, where am i adding a dev tag then?

illusional: On deploy

illusional (approved):

PR #65, opened 2022-12-13
Add CI step to build special case `cpg_workflows` from production_pipelines repo

vladsavelyev:


lgruen (approved):
That's actually much nicer than I thought it would be. Awesome work!

PR #66, opened 2022-12-13
Install DRAGMAP from conda instead of installing from source

lgruen:
This seems to work fine for 1.3.0 by now.


lgruen:
Causes a segfault: https://batch.hail.populationgenomics.org.au/batches/396768/jobs/1

PR #70, opened 2022-12-14
Reformat VEP Dockerfile

lgruen:
Functional no-op


MattWellie (approved):
My only hesitancy with going to a single RUN was that there would be no parallelised pulling of the container layers from the artifactory (... and while debugging the latter commands, having to re-gen the whole image each time would have been awful). 

I don't think it's likely to be an issue though

PR #73, opened 2023-01-31
add docker image for saige-qtl

annacuomo:
Docker image to install [SAIGE-qtl edition](https://github.com/weizhou0/qtl), an extension of [SAIGE](https://github.com/saigegit/SAIGE) for QTL mapping.

Since it is seemingly working now (after the latest fix) - this should be ready to be merged now?


images/saige-qtl/Dockerfile
```diff
@@ -0,0 +1,13 @@
+FROM australia-southeast1-docker.pkg.dev/analysis-runner/images/driver-r:1.1
```
lgruen, line 1:
```suggestion
FROM australia-southeast1-docker.pkg.dev/analysis-runner/images/driver-r:1.4
```

images/saige-qtl/Dockerfile
```diff
@@ -0,0 +1,13 @@
...
+
+SHELL ["/bin/bash", "-eo", "pipefail", "-c"]
+RUN conda create -n RSAIGE r-essentials r-base=4.1.2 python=3.10 \
```
lgruen, line 4: `micromamba` will be much faster to resolve things than `conda`.

Can you please change this to use `micromamba install -y --prefix $MAMBA_ROOT_PREFIX ...` (see https://github.com/populationgenomics/analysis-runner/blob/24673e40894262513ab2bc79c5e9ac0996921cf2/driver/Dockerfile.r#L14-L24)? You don't need to create a new conda environment either, just install it in the base environment (since this is isolated in a Docker image anyway). Also best not to chain multiple `install` operations, but instead install everything at once. This way, the dependencies only need to be resolved once.

lgruen (changes requested):

images/saige-qtl/Dockerfile
```diff
@@ -0,0 +1,18 @@
+FROM australia-southeast1-docker.pkg.dev/analysis-runner/images/driver-r:1.4
+
+ENV MAMBA_ROOT_PREFIX /root/micromamba
+
+SHELL ["/bin/bash", "-eo", "pipefail", "-c"]
+RUN wget -qO- https://api.anaconda.org/download/conda-forge/micromamba/0.8.2/linux-64/micromamba-0.8.2-he9b6cbd_0.tar.bz2 | tar -xvj -C /usr/local bin/micromamba && \
+    mkdir $MAMBA_ROOT_PREFIX && \
```
lgruen, line 7: You don't need to install `micromamba`, as you're already using the R driver image as a base.

lgruen (changes requested):

images/saige-qtl/Dockerfile
```diff
@@ -0,0 +1,16 @@
+FROM australia-southeast1-docker.pkg.dev/analysis-runner/images/driver-r:1.4
+
+ENV MAMBA_ROOT_PREFIX /root/micromamba
+
```
lgruen, line 4:
```suggestion
```

images/saige-qtl/Dockerfile
```diff
@@ -0,0 +1,16 @@
...
+SHELL ["/bin/bash", "-eo", "pipefail", "-c"]
+RUN micromamba install -y --prefix $MAMBA_ROOT_PREFIX \
+        -c cpg -c bioconda -c conda-forge \
+        r-essentials r-base=4.1.2 python=3.10 \
+        anaconda cmake \
```
lgruen, line 9: That's probably already installed through the base as well, in addition to a few of the things below?

images/saige-qtl/Dockerfile
```diff
@@ -0,0 +1,16 @@
...
+        gettext lapack r-matrix \
+        r r-rcpp r-rcpparmadillo r-data.table r-bh r-matrix \
+        r-spatest r-rcppeigen r-devtools r-skat \
+        r-rcppparallel r-optparse boost openblas r-rhpcblasctl r-metaskat \
+        r-skat r-qlcmatrix r-rsqlite \
+    rm -r /root/micromamba/pkgs \
+    pip3 install cget click
```
lgruen, line 16: Did you mean to concatenate these with `&&` at the end of the lines?

images/saige-qtl/Dockerfile
annacuomo: yes good point, i can try and remove one by one to check? Or is there a better way to make sure?

lgruen: I don't know a great way for conda unfortunately. The other approach is to build it up in reverse: only start with what you know you truly need and then add more packages when you encounter remaining dependency problems.

images/saige-qtl/Dockerfile
```diff
@@ -0,0 +1,14 @@
...
+        -c cpg -c bioconda -c conda-forge \
+        r-essentials r-base=4.1.2 python=3.10 \
+        anaconda cmake \
+        gettext lapack r-matrix \
+        r r-rcpp r-rcpparmadillo r-data.table r-bh r-matrix \
+        r-spatest r-rcppeigen r-devtools r-skat \
+        r-rcppparallel r-optparse boost openblas r-rhpcblasctl r-metaskat \
+        r-skat r-qlcmatrix r-rsqlite && \
+    rm -r /root/micromamba/pkgs && \
+    pip3 install cget click
```
lgruen, line 14: Do you need `cget`?

images/saige-qtl/Dockerfile
annacuomo: I am installing everything recommended [here](https://github.com/saigegit/SAIGE/blob/main/conda_env/createCondaEnvSAIGE_steps.txt) - linked from https://saigegit.github.io/SAIGE-doc/docs/Installation_conda.html

images/saige-qtl/Dockerfile
annacuomo: yes sorry my bad

images/saige-qtl/Dockerfile
annacuomo: again I assume she wouldn't have [listed it](https://github.com/saigegit/SAIGE/blob/main/conda_env/createCondaEnvSAIGE_steps.txt) unless it wasn't necessary, but I guess I could try running the actual package with a reduced set of dependencies and see?

images/saige-qtl/Dockerfile
annacuomo: yes i see that now, thank you!

images/saige-qtl/Dockerfile
annacuomo: yes should be done now!

images/saige-qtl/Dockerfile
lgruen: Ah, got it! `anaconda` is not a package though, it's the channel (the parameter after `-c`).

annacuomo: oops my bad! moved up now

images/saige-qtl/Dockerfile
```diff
@@ -0,0 +1,13 @@
+FROM australia-southeast1-docker.pkg.dev/analysis-runner/images/driver-r:1.4
+
+SHELL ["/bin/bash", "-eo", "pipefail", "-c"]
+RUN micromamba install -y --prefix $MAMBA_ROOT_PREFIX \
+        -c cpg -c bioconda -c conda-forge -c anaconda \
+        r-essentials r-base=4.1.2 python=3.10 \
+        cmake gettext lapack r-matrix \
+        r r-rcpp r-rcpparmadillo r-data.table r-bh r-matrix \
```
lgruen, line 8: `r` is again just the channel. You'll want to move all the `-c` parameters to the top.

images/saige-qtl/Dockerfile
lgruen: It's not a huge deal, I was just surprised to see it listed. I'd be surprised if it was actually necessary.

images/saige-qtl/Dockerfile
```diff
@@ -0,0 +1,13 @@
+FROM australia-southeast1-docker.pkg.dev/analysis-runner/images/driver-r:1.4
+
+SHELL ["/bin/bash", "-eo", "pipefail", "-c"]
+RUN micromamba install -y --prefix $MAMBA_ROOT_PREFIX \
+        -c cpg -c bioconda -c conda-forge -c anaconda \
+        r-essentials r-base=4.1.2 python=3.10 \
```
lgruen, line 6: All of this should already be installed.

```suggestion
```

images/saige-qtl/Dockerfile
annacuomo: ok! It's building again now, will see if it works.
I just realised that this is actually still only installing all of the dependencies, not the package itself? Can I do that from source (from [here](https://saigegit.github.io/SAIGE-doc/docs/Installation_conda.html)) within the Dockerfile?:

>   src_branch=main
    repo_src_url=https://github.com/saigegit/SAIGE
    git clone --depth 1 -b $src_branch $repo_src_url
    Rscript ./SAIGE/extdata/install_packages.R
    R CMD INSTALL --library=path_to_final_SAIGE_library SAIGE

lgruen: Yeah, it's fairly annoying they neither publish a package nor version their releases. So we can't even easily build this reproducibly without hardcoding a GitHub SHA here...

I'll change this slightly and also integrate it with the top-level versioning of this repo.

lgruen (approved):
I've added a commit to fix the build, but I'm not sure how to test that the installation actually worked.

Before merging you might want to update the PR description.

images/saige-qtl/Dockerfile
annacuomo: ok! To test, would we be able to just run, e.g.,:

```Rscript createSparseGRM.R --help```

to make sure the various functions / scripts have been installed properly?

File: images/saige-qtl/Dockerfile [hunk 10]
```diff
@@ -0,0 +1,39 @@
...
+    chmod a+x /usr/local/bin/step3_LDmat.R && \
+    chmod a+x /usr/local/bin/createSparseGRM.R
+
+RUN createSparseGRM.R  --help && \
+    step1_fitNULLGLMM_qtl.R --help && \
+    step2_tests_qtl.R --help && \
+    step3_LDmat.R --help
+
+RUN apt-get update
+RUN apt-get install time
```
lgruen, line 39:
```suggestion
    R CMD INSTALL . && \
    mv extdata/step1_fitNULLGLMM_qtl.R \
          extdata/step2_tests_qtl.R \
          extdata/step3_LDmat.R \
          extdata/createSparseGRM.R /usr/local/bin/ && \
    chmod a+x /usr/local/bin/step1_fitNULLGLMM_qtl.R \
                        /usr/local/bin/step2_tests_qtl.R \
                        /usr/local/bin/step3_LDmat.R \
                        /usr/local/bin/createSparseGRM.R && \
    cd .. && \
    rm -r qtl
```

lgruen (changes requested):

lgruen (approved):

lgruen (approved):

PR #77, opened 2023-02-17
multistage version, image size from ~400MB to 80MB

MattWellie:
Posting a multistage build alternative, after discussions on reducing image sizes to reduce overall job setup/image pulling times.

Would appreciate feed back on whether this is a good idea, and if so whether we should start trying to finesse our more frequently used images

### DIFF
--- a/images/bgt/Dockerfile
+++ b/images/bgt/Dockerfile
@@ -1,4 +1,4 @@
-FROM ubuntu:20.04
+FROM ubuntu:20.04 AS compiler
 ARG DEBIAN_FRONTEND=noninteractive
 ARG VERSION=${VERSION:-v1.0}
 
@@ -28,4 +28,4 @@ RUN apt-get update -y && apt-get install -y --no-install-recommends -qqy \
 
 FROM ubuntu:20.04
 ARG DEBIAN_FRONTEND=noninteractive
-COPY --from=0 /bgt/bgt /usr/local/bin/bgt
+COPY --from=compiler /bgt/bgt /usr/local/bin/bgt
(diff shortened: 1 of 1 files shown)


images/bgt/Dockerfile
```diff
@@ -20,11 +20,12 @@ RUN apt-get update -y && apt-get install -y --no-install-recommends -qqy \
...
-    && chmod a+x /usr/local/bin/bgt \
-    && cd .. \
-    && rm -rf bgt
\ No newline at end of file
+    && chmod a+x bgt
+
+
+FROM ubuntu:20.04
+ARG DEBIAN_FRONTEND=noninteractive
+COPY --from=0 /bgt/bgt /usr/local/bin/bgt
```
lgruen, line 31: You can also name stages to make this a bit more readable.

lgruen (approved):
This is nice, but only works for completely statically linked libraries. E.g. even something like SSL certificates can be problematic and only shows up once you try to establish a connection. There are some tools to work around this (e.g. what I've used here: https://github.com/populationgenomics/cuKING/blob/1d29d72a184da33cf2500f8b537ae56ce66a7e96/Dockerfile#L144), but it typically requires careful testing.

Generally I'd love to have smaller images (particularly for the default driver!), but as long as we rely on `gcloud` from the Google SDK, it's all a bit hopeless ;). Also some pipelines want to install additional packages (e.g. for R) that require compilers being installed, so the super slim base images don't work well for that either.

For our use case, the lazy loading of Docker image blocks (https://github.com/containerd/stargz-snapshotter) is overall more promising, I believe. So you can have large images, but you only download the small fraction that you actually need. The Hail team is looking into that, but they want to finish migrating to Podman first.

images/bgt/Dockerfile
MattWellie: Agreed all over, I think this made for a decent template because it's an isolated tool. That will likely apply to a couple of other images, but not many. The Google SDK is going to be quite limiting, but I'm still planning to play around with a different base image e.g. for AIP - there's no reason for that to be a 2.6GB image 😱

PR #78, opened 2023-02-20
skinny python base

MattWellie:
First crack at making a skinny Python base to build more niche python images on, without all the cumbersome GCloud SDK weight. I've smashed together the 2 Dockerfiles for `Base` and `Hail` from analysis-runner, and kept libraries likely to be broadly useful to CPG jobs - utils, workflows, hail, cloudpathlib, sample-metadata, gcsfs. That could probably be reduced further, but I'm having problems installing Hail, which feels like a core requirement. 

My first attempt at building an image from this got stuck for over an hour during the make install of Hail, about 10% through a Gradle download. My internet connection has been awful lately, so it could have been caused by disconnection during the build, but I tried a while back and gave up due to the same inability to test-build locally. 

Have you got any experience with building this locally, and is it usually a PITA?


illusional:
What's the final image size? Is it worth the effort to only simply exclude gcloud?

MattWellie:
@illusional this is piggybacking on https://github.com/hail-is/hail/pull/12526, removal of the gcloud SDK cuts over 1GB. AFAIK that's over half of the Driver content and isn't used in most jobs.

illusional:
Makes sense, given we would want to keep this image very-up-to-date, would it be worth moving this to the analysis-runner, that redeploys this image frequently, and we could have a multistage build that throws out an additional image with the gcloud layer?

lgruen:
One issue we'd run into with this is that the analysis-runner currently _always_ runs `gcloud auth` as part of the initial invocation: https://github.com/populationgenomics/analysis-runner/blob/32b9906d973e837d153d850b0207c77011590708/analysis_runner/git.py#L153-L156

A lot of the simple `gsutil cp` type scripts rely on this. Maybe we just need to make this more explicit going forward, but it would be a breaking change. Similarly, some of the production pipelines jobs run `GCS_OAUTH_TOKEN=$(gcloud auth application-default print-access-token)` for `samtools` etc., so unfortunately we can't easily have a separate `gcloud-sdk` image only intended for bash scripts either.

> I've smashed together the 2 Dockerfiles for Base and Hail from analysis-runner

How would you feel about removing the SDK from `base` instead? We'd still need to resolve the above problems somehow, but it might be a little cleaner?

MattWellie:
Definitely open to it, I just wanted to try a few build variations and see what the resulting image size was... so far I've not been able to build anything with hail so I wondered if there's some kind of trick to it 😓

lgruen:
There's no trick to it, it just takes a while to build locally on a Mac. However, it's _way_ faster to experiment with long Docker builds simply on a Linux VM (e.g. use could use your `dev` GCP playground project for it).

MattWellie:
@lgruen I'd like to revisit this, but my failure to build it was a stumbling block - can we catch up on how to run cloud builds?
Either that or we can merge this PR, get github CI to run the build, then I can pull that locally as a base and experiment with building on top of it

lgruen:
Yeah, happy to! Do you want to chat on Zoom some time today? I usually either use `gcloud builds` or experiment directly on a Linux VM for more interactive / iterative `Dockerfile` writing.

PR #79, opened 2023-03-02
Add bwa

cassimons:


images/bwa/Dockerfile
```diff
@@ -0,0 +1,15 @@
...
+
+RUN apt-get update && apt-get install -y \
+        git wget bash bzip2 zip build-essential libboost-all-dev libz-dev && \
+    rm -r /var/lib/apt/lists/* && \
+    rm -r /var/cache/apt/* && \
+    wget -qO- https://api.anaconda.org/download/conda-forge/micromamba/0.8.2/linux-64/micromamba-0.8.2-he9b6cbd_0.tar.bz2 | tar -xvj -C /usr/local bin/micromamba && \
+    mkdir ${MAMBA_ROOT_PREFIX} && \
+    micromamba install -y --prefix ${MAMBA_ROOT_PREFIX} -c bioconda -c conda-forge \
+        bazam biobambam google-cloud-sdk samtools bwa \
+    && rm -r /root/micromamba/pkgs
```
lgruen, line 15:
```suggestion
        bazam biobambam google-cloud-sdk samtools bwa && \
    rm -r /root/micromamba/pkgs
```

lgruen (approved):
Might be useful to give a bit of context of why we want to resurrect this image, given that we've broadly switched to DRAGMAP.

images/bwa/Dockerfile
cassimons: The mito pipeline requires the realignment of reads to the mito genome. Broad use BWA for this. While we could consider doing this with dragmap I want to start from a functionally equivalent pipeline before swapping tools out.

PR #88, opened 2023-06-15
Container with BaseSpace CLI installed

EddieLF:
A container image based on the AR driver base, which just installs BaseSpace CLI to the $HOME/bin/bs directory

See https://github.com/populationgenomics/transfer-private/pull/7 for intended use


images/basespace/Dockerfile
```diff
@@ -0,0 +1,9 @@
+FROM australia-southeast1-docker.pkg.dev/analysis-runner/images/driver-base:1.4
+
+SHELL ["/bin/bash", "-eo", "pipefail", "-c"]
```
illusional, line 3: Hail doesn't use the SHELL command, it inserts a script into the container, and runs it instead (so you could omit this).

illusional (approved):

PR #90, opened 2023-06-16
Use version 1.5.3 for BaseSpace CLI and add to images.toml

EddieLF:
My image didnt get pushed to the cpg-common Artifact registry, I think because I hadn't added it to the images toml. This should fix that.


images/basespace/Dockerfile
```diff
@@ -1,6 +1,11 @@
...
+ARG VERSION=${VERSION:-1.5.3}
+
 RUN apt-get update && \
     mkdir $HOME/bin && \
-    wget "https://launch.basespace.illumina.com/CLI/latest/amd64-linux/bs" -O $HOME/bin/bs && \
-    chmod u+x $HOME/bin/bs
+    wget "https://launch.basespace.illumina.com/CLI/${VERSION}/amd64-linux/bs" -O $HOME/bin/bs && \
+    chmod u+x $HOME/bin/bs && \
+    pip3 install cpg_utils && \
+    pip3 install cpg_workflows && \
```
illusional, line 10: What are you doing in this image? Adding cpg_workflows suggests that you might be reusing this image outside the regular basespace flow

EddieLF: `cpg_workflows` was previously in this image to use the `get_batch()` function in scripts in the transfer-private repo. 

I'm removing `cpg_workflows` from the image and changing the scripts to import `get_batch` from `cpg_utils` instead. Once I can merge these changes into main, I will rebuild and deploy this image to use with the updated transfer scripts.

illusional: Do you have a script where you're using this?
If it's just the copying bit, I feel like this image could be a bit lighter, you don't necessarily want to use this for the "driver" image when running through the analysis-runner, you can just specify it in your workflow as the image.

illusional (approved):

PR #92, opened 2023-07-15
Add Mitoreport

cassimons:
Mitoreport is a tool developed by VCGS to generate standalone HTML reports for the interpretation of mito variants. We will use this in the RD mito pipeline.


vivbak (approved):
Hey @cassimons, assume you meant to ask for the review rather than assign :)

cassimons:
@vivbak sorry, yes. And meant to say it could wait till tomorrow as well.

PR #93, opened 2023-07-28
Add a fastp image

cassimons:


images/fastp/Dockerfile
```diff
@@ -0,0 +1,14 @@
+FROM debian:buster-slim
+
+ENV MAMBA_ROOT_PREFIX /root/micromamba
+ENV PATH $MAMBA_ROOT_PREFIX/bin:$PATH
+ARG VERSION=${VERSION:-0.23.4}
+
+RUN apt-get update && apt-get install -y git wget bash bzip2 zip && \
+    rm -r /var/lib/apt/lists/* && \
+    rm -r /var/cache/apt/* && \
+    wget -qO- https://api.anaconda.org/download/conda-forge/micromamba/0.8.2/linux-64/micromamba-0.8.2-he9b6cbd_0.tar.bz2 | tar -xvj -C /usr/local bin/micromamba && \
```
jmarshall, line 10: Current version of micromamba is 1.4.9, but it's probably better to leave it the same version as in our other images, as you've done.

jmarshall (approved):
Looks good. I'd suggest basing on a more current Debian version as per comment, but happy either way.

jmarshall:
There is this thread where I was thinking about refactoring our micromamba-using images — https://centrepopgen.slack.com/archives/C03FZL2EF24/p1684280656650199 — but I haven't done anything about that, so one more micromamba image is fine.

images/fastp/Dockerfile
```diff
@@ -0,0 +1,14 @@
+FROM debian:buster-slim
```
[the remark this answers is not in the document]
cassimons: I thinking this was getting a bit long in the tooth - bumped to bookworm

images/fastp/Dockerfile
cassimons: micromamba is superfluous here so I ditched it

jmarshall: Fair enough.

What's $PWD at this point? Maybe `/`?

Might be better to put it in /usr/bin or /usr/local/bin -- somewhere already on $PATH.

cassimons: Thanks, yea discovered that when I tested it - linked to /usr/local/bin now

PR #94, opened 2023-08-03
ExpansionHunterv5 - with modifications from Ben W

hopedisastro:
Modification includes changing an error into a warning, to solve the issue here: https://centrepopgen.slack.com/archives/C030X7WGFCL/p1690439533965099


images.toml
```diff
@@ -28,3 +28,4 @@ stripy = '36a3fec9586982c8226dc4bdabd5c9108adcb0c1'
 gangstr = '2.5.0'
 hipstr = 'v0.6.2'
 bgt = 'v1.0'
+expansionhunter_bw2 = '5.0.0-bw2'
```
jmarshall, line 31: It would be best to alphabetise this alongside the other `expansionhunter` entry.

images/expansionhunter_bw2/Dockerfile
```diff
@@ -0,0 +1,30 @@
...
+    libbz2-dev \
+    libcurl4-openssl-dev \
+    liblzma-dev \
+    libssl-dev \
+    zlib1g-dev
+
+# Clone the repository, jump into it
+# build the ExpansionHunter executable, +x, move to /usr/bin
+RUN git clone https://github.com/bw2/ExpansionHunter && \
+    cd ExpansionHunter && \
```
jmarshall, line 21: At the moment, this just uses **master** from bw2's fork. This may move in future, so for reproducibility we need to identify the particular commit we currently want. So please checkout a commit explicitly here (between the `cd` and `mkdir build`):

```
    git checkout $VERSION && \
```

The commit to use is the current head of bw2's **master**, which is b85cba004fba4e0223e632fae576bc7fad0da804.

And then use that instead of `5.0.0-bw2` in the `ARG VERSION` line above and in _images.toml_.

This image needs a particular untagged commit hash, so the version number bits should look like the images for _haplocheckcli_, _saige-qtl_, and _stripy_.

images/expansionhunter_bw2/Dockerfile
```diff
@@ -0,0 +1,30 @@
...
+    mkdir build && \
+    cd build && \
+    cmake .. && \
+    make && \
+    chmod +x /ExpansionHunter/build/install/bin/ExpansionHunter && \
+    mv /ExpansionHunter/build/install/bin/ExpansionHunter /usr/bin
+
+# Set the entry point for the Docker container
+ENTRYPOINT ["ExpansionHunter"]
```
jmarshall, line 30: Is there a reason you're setting an ENTRYPOINT here? Most of our images don't, and it seems like it would be best to set this one up like the plain expansionhunter image, which also doesn't.

images/expansionhunter_bw2/Dockerfile
```diff
@@ -0,0 +1,30 @@
...
+# Clone the repository, jump into it
+# build the ExpansionHunter executable, +x, move to /usr/bin
+RUN git clone https://github.com/bw2/ExpansionHunter && \
+    cd ExpansionHunter && \
+    mkdir build && \
+    cd build && \
+    cmake .. && \
+    make && \
+    chmod +x /ExpansionHunter/build/install/bin/ExpansionHunter && \
+    mv /ExpansionHunter/build/install/bin/ExpansionHunter /usr/bin
```
jmarshall, line 27: It'd be good to add some cleanup here, mainly delete the build directory (1.3 Gb!):

```
rm -rf /ExpansionHunter
```

images/expansionhunter_bw2/Dockerfile
```diff
@@ -0,0 +1,30 @@
...
+RUN apt-get update && \
+    apt-get install -y \
+    cmake \
+    g++ \
+    git \
+    libbz2-dev \
+    libcurl4-openssl-dev \
+    liblzma-dev \
+    libssl-dev \
+    zlib1g-dev
```
jmarshall, line 16: It's a good idea to clean out the apt caches after all this installation, as after here they're just extraneous bloat in the docker image:

```suggestion
    zlib1g-dev \
    && \
    apt-get clean -y && \
    rm -rf /var/lib/apt/lists/*
```

jmarshall (changes requested):
This mostly looks good.

A few comments; the main one that needs to be addressed is about being explicit that we're building an image from explicitly commit b85cba004fba4e0223e632fae576bc7fad0da804.

This will want to be specified in _images.toml_ and the `VERSION` bits in _Dockerfile_ updated as per the review comment.

jmarshall:
I also verified that the C++ code does indeed contain the modification from error about flanking Ns to a warning. This looks like it has been changed in bw2/ExpansionHunter@7af56b7246cf79047f9edc244ecb144ac87c834f.

hopedisastro:
Thanks John for reviewing this PR! I think I've incorporated all of your comments.

jmarshall (approved):

PR #99, opened 2023-09-20
Adds a new VEP builder

MattWellie:
Burns down the previous process and starts from scratch:

- Build on the released Docker image from Ensembl (Trusting them to build their own tool...)
- Sets the user back to root so we can actually use our own image without user permission errors
- Pulling the GRCh38 version of LOFTEE and overwriting any plugins that existed previously (default = GRCh37 which is bad for us)
- Pulling in some small reference files for plugins that don't have this data packaged by default 
  - utr_annotator_38.txt: 2.3MB
  - pli_scores.txt: 0.5MB
  - loftool_110_scores.txt: 233KB
- Pulling in the AlphaMissense plugin (hopefully should be bundled by default soon, it was only released yesterday...)

This approach scraps some stuff from the alternative `images/vep/Dockerfile` approach:
- Don't use conda/miniconda/anyconda, just rely on a pre installed version
- Don't bother installing gcsfuse - our workflows rely on hail batch to fuse a read only container as a Docker volume mount (AFAIK) - the container itself doesn't need to gcsfuse anything?? (this may be incorrect)
- Don't bother installing google-cloud-cli - similar reasons, I think Hail Batch does the I/O for our implementation of VEP, we don't push anything directly to GCP (again, this could be a little misguided)

My local size for this completed image: ~850MB

### DIFF
--- a/images/ensembl-vep/Dockerfile
+++ b/images/ensembl-vep/Dockerfile
@@ -1,0 +1,63 @@
+ARG VERSION=${VERSION:-release_110.1}
+
+FROM ensemblorg/ensembl-vep:${VERSION}
+
+USER root
+
+## Copying in a couple of small data files, and a plugin or two
+
+# AlphaMissense
+# print AlphaMissense scores and predictions
+# only report results for the transcripts in the AlphaMissense prediction
+# ./vep -i variations.vcf --plugin AlphaMissense,file=/full/path/to/file.tsv.gz,transcript_match=1
+# file path = gs://dm_alphamissense/AlphaMissense_aa_substitutions.tsv.gz
+ENV AlphaMissensePlugin=${VEP_DIR_PLUGINS}/AlphaMissense.pm
+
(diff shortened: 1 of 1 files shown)


cassimons:
If we are not using gcsfuse, how/where are we maintaining a central VEP cache? I assume this is not in the image or it would be huge?

MattWellie:
As I understand it:
- VEP Cache is stored in GCS
- Hail Batch mounts this as a volume
- VEP (this container) won't need any special libraries to interact with it

The original version of this VEP image installed gcs-fuse internally, but doesn't initiate any fuse's
@hopedisastro has been working on some [SV-Workflows](https://github.com/populationgenomics/sv-workflows/blob/main/str/runners/str_iterative_hipstr_runner.py#L32), some of which use GCS-fuse to mount in read-only data. The [image](https://github.com/populationgenomics/images/blob/main/images/hipstr/Dockerfile) for those jobs has no Google client libraries installed

cassimons (approved):

PR #101, opened 2023-10-01
Update Dockerfile

hopedisastro:
I am getting a pickling error: https://batch.hail.populationgenomics.org.au/batches/428058/jobs/1, despite not doing any pickling/unpickling in my script. (https://github.com/SeldonIO/seldon-core/issues/4315) suggests that downgrading the pandas version may solve the pickling issue

### DIFF
--- a/images/cellregmap/Dockerfile
+++ b/images/cellregmap/Dockerfile
@@ -19,7 +19,7 @@ RUN apt-get update && \
         click==8.0.4 \
         limix==3.0.4 \
         numpy==1.21.2 \
-        pandas==1.3.4 \
+        pandas==1.2.3 \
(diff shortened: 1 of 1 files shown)


hopedisastro:
I tested out the functions in a jupyter notebook yesterday. It worked fine with pandas==1.3.5 (I just checked the version), so now I'm not sure if the pandas version is the issue or if it's a larger cloud/image related issue.

hopedisastro:
Ah, never mind. I've isolated the issue to the Hail batch python job function: https://hail.is/docs/batch/api/utils/hailtop.batch.docker.build_python_image.html#hailtop.batch.docker.build_python_image

PR #103, opened 2023-10-02
Add FastQE

michael-harper:
Adding FastQE for onboarding documentation purposes


MattWellie:
Is there a reason for doing a micromamba installation in a naked OS image instead of starting with a python base image and installing with PIP?

Likewise is there a reason for google cloud sdk to be installed?

Both of these are bad CPG docker habits rather than good practice. Ideally you want a single tool installed in the simplest way possible.

michael-harper:
> Is there a reason for doing a micromamba installation in a naked OS image instead of starting with a python base image and installing with PIP?
> 
> Likewise is there a reason for google cloud sdk to be installed?
> 
> Both of these are bad CPG docker habits rather than good practice. Ideally you want a single tool installed in the simplest way possible.

FastQE is going to need access to the files on GCS and so will need google cloud sdk (is there another way to approach this???). I tried not using micromamba and just using the [google cloud sdk docker image](https://hub.docker.com/r/google/cloud-sdk/) but that was still around 1.3Gb I believe. The above solution manages to install fastqe and google cloud sdk straight-forwardly and is only marginally bigger at 1.4Gb.

MattWellie (changes requested):
I feel strongly that this image should be a tool-only image, with a minimal installation of a single tool. Currently this is modelled on other CPG Dockerfiles which don't follow good platform-independent practice.

cassimons:
Hey Michael, just piping in here that given this image is going to be the first image that new starters interact with I think we want to make sure it follows current best practice even if the practical difference in this case is limited. 

> FastQE is going to need access to the files on GCS and so will need google cloud SDK

Are you reading directly from buckets, or are you intending to use standard Batch file localisation? If so, you should not need the google libraries (and is probably the better method to use as a demonstration).

PR #104, opened 2023-10-06
Add Perl lib VEP-110

MattWellie:
Indexed data files vastly improve speed of VEP annotation, but this requires the installation of Tabix.

n.b. I have already manually built and pushed from this branch, so this is not a pressing PR

### DIFF
--- a/images/ensembl-vep/Dockerfile
+++ b/images/ensembl-vep/Dockerfile
@@ -71,6 +71,7 @@ RUN export DEBIAN_FRONTEND=noninteractive \
 	&& make \
 	&& make install \
 	&& cd .. \
-	&& rm -rf samtools-${HTS_VERSION}.tar.bz2 samtools-${HTS_VERSION}
+	&& rm -rf samtools-${HTS_VERSION}.tar.bz2 samtools-${HTS_VERSION} \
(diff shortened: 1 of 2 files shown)


jmarshall:
Probably I'm missing something, but this image is based on ensemblorg/ensembl-vep (which is based on Ubuntu 18.04, yay…), and that already has `bgzip` and `tabix` installed. They're quite old (1.9) but the compression and indexing haven't changed much if at all.

MattWellie:
> Probably I'm missing something, but this image is based on ensemblorg/ensembl-vep (which is based on Ubuntu 18.04, yay…), and that already has `bgzip` and `tabix` installed. They're quite old (1.9) but the compression and indexing haven't changed much if at all.

Yeah that's entirely fair. Initially this was to install the `Ensembl::XS` perl lib which is identified [here](https://asia.ensembl.org/info/docs/tools/vep/script/vep_other.html) as one of the many ways to improve runtime. I've pulled the resulting image and actually the tabix version is still recognised as 1.9, so this additional build is useless.

I'll scrape this back down to just adding that perl lib

MattWellie:
I also realised that this is adding a second dockerfile which shouldn't see the light of day

PR #106, opened 2023-10-08
hail 0.2.124 image

hopedisastro:
In response to this fix by the Hail team: https://centrepopgen.slack.com/archives/C018KFBCR1C/p1694041536595649?thread_ts=1693545507.704539&cid=C018KFBCR1C, I need an updated hail version to run on analysis-runner to see if this fixes the LD prune recursion issue.


hopedisastro:
I think I've used the driver image that should work on analysis-runner, but let me know if that's not the case

images/hail/Dockerfile
```diff
@@ -0,0 +1,11 @@
+FROM australia-southeast1-docker.pkg.dev/analysis-runner/images/driver:a15ed15c0ca0b7304e0ceca1b873d734720c5d7a-hail-abaf45c99b32250dfc55a1b26a6a0635760e0a99
+
+# ARG VERSION=${VERSION:-0.2.124}
```
MattWellie, line 3: Should this be commented out?

MattWellie:
There's an issue with the latest version(s?) of Hail which are causing problems starting Hail Batches. Will you be using this image to run a single job, or to plan a batch?

PR #108, opened 2023-10-10
Modify docker builder workflow

MattWellie:
Modify the target artifactory depending on branch (main vs. non-main) status


jmarshall:
You could do some trickery like I did here

https://github.com/pysam-developers/pysam/blob/08de757a39ab42e867fa188faa860a5fd0b648c1/.github/workflows/release.yaml#L84

to write the step once instead of twice. But to be honest, for one step clearer is better :smile:

MattWellie:
> You could do some trickery like I did here

I like the trickery, but I also like that the two-step will have one active stage in the github actions and one skipped, each with identifiable names. IMO that makes it easier at a glance to see where the images are ending up

illusional (approved):
Awesome, nice trio of pull-requests.
I'll update all the other branches on this repo so they get this change :)

PR #116, opened 2023-11-02
Add samtools to vep 110

MattWellie:
I previously submitted this PR: https://github.com/populationgenomics/images/pull/104

Long story short - I was adding Samtools to the image, but in the PR I incorrectly stated I was installing Tabix. John correctly identified that the image would already had Tabix installed. I had a brain fart and closed the PR, opening a new one without Samtools... Sigh.

This error has surfaced now that the image has moved to a new name (`ensembl-vep` -> `vep_110`). I built `ensembl-vep` from the branch, and `vep_110` has been built from main (no Samtools).

### DIFF
--- a/images/vep_110/Dockerfile
+++ b/images/vep_110/Dockerfile
@@ -1,6 +1,7 @@
 ARG VERSION=${VERSION:-release_110.1}
 
 FROM ensemblorg/ensembl-vep:${VERSION}
+ARG HTS_VERSION=${HTS_VERSION:-1.18}
 
 USER root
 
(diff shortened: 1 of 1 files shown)


MattWellie:
FYI @EddieLF this change will be required for new annotation jobs (maybe not _required_, but the runtime will be way longer without it)

cassimons (approved):

jmarshall:
Dumb/curious question: What samtools command(s) does it shell out to?

MattWellie:
> Dumb/curious question: What samtools command(s) does it shell out to?

It's in the LoFtee plugin: https://github.com/konradjk/loftee/blob/grch38/LoF.pm#L539

The rest of the annotation would be applied as expected, but LOFTEE would fail, and the failure-catching loop causes a noticeable slowdown

jmarshall:
You would think there would already be Bio-DB-HTS or some other Perl module installed that can query an indexed reference…

Anyway: this comes under hmmmfph software. Thanks for the explanation, and fair enough.

For just this we could just install Debian's samtools. But since you've already written the build script, fair enough again!

PR #117, opened 2023-11-06
Remove Prod-Pipes PAT

MattWellie:
Repo is no longer private, and this expired token is crashing the deploy workflow
Also bumps the gcloud setup workflow which is available at `V1`
https://github.com/marketplace/actions/set-up-gcloud-cloud-sdk-environment

### DIFF
--- a/.github/workflows/deploy_config.yaml
+++ b/.github/workflows/deploy_config.yaml
@@ -87,8 +87,6 @@ jobs:
         ref: "main"
         path: "production-pipelines"
         submodules: recursive
-        # Comment this line out to revert to using the default $GITHUB_TOKEN.
-        token: ${{ secrets.PRODUCTION_PIPELINES_RAW_REPO_TOKEN }}
 
     - id: get_version
       run: |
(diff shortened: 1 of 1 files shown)


MattWellie:
useful thread https://github.com/actions/checkout/issues/664

.github/workflows/deploy_config.yaml
```diff
@@ -87,8 +87,6 @@ jobs:
         ref: "main"
         path: "production-pipelines"
         submodules: recursive
-        # Comment this line out to revert to using the default $GITHUB_TOKEN.
-        token: ${{ secrets.PRODUCTION_PIPELINES_RAW_REPO_TOKEN }}
```
jmarshall, line 91: I think this is the one I was expecting to stop working when this token expired. The others can fall back to unauthenticated access to now-public repositories, but not this one that's explicitly setting an action's `token` parameter.

So that this remains a hint for next time we have to make that repo private (hopefully never happens, but it's hard to make predictions about the future), I'd rather this was just commented out and the comment I guess updated to something like _Uncomment this line and set up a PAT if production-pipelines needs to be taken private_.

jmarshall (approved):
Agree that this `token` line needs to be deactivated and the `setup-gcloud@v1` change is good. A comment for your consideration about how to keep the `token` documentation for the future…

.github/workflows/deploy_config.yaml
MattWellie: fair comment, I'll do that instead!

PR #118, opened 2023-11-16
Update actions to use google-github-actions/auth@v1 and fix input syntax

EddieLF:
So it looks like the new version of `google-github-actions/setup-gcloud@v1` action no longer uses `service_account_key` as input. See [this failed deploy](https://github.com/populationgenomics/images/actions/runs/6886508598/job/18732356675), under the gcloud setup header. 
```
Warning: Unexpected input(s) 'service_account_key', valid inputs are ['skip_install', 'version', 'project_id', 'install_components']
...
Warning: No authentication found for gcloud, authenticate with `google-github-actions/auth`.
```
Which is leading to a fail in the `deploy config toml` step.
```
ERROR: (gcloud.storage.cp) You do not currently have an active account selected.
Please run:

  $ gcloud auth login

to obtain new credentials.
```

From the actions repo, it looks like they now require use of the `google-github-actions/auth` action to first auth with the `service_account_key` (now called `credentials_json`) before running the gcloud setup.

Also updates the `.github/workflows/deploy_container.yaml` to use the `@v1` version of the gcloud setup as well.

### DIFF
--- a/.github/workflows/deploy_config.yaml
+++ b/.github/workflows/deploy_config.yaml
@@ -52,10 +52,16 @@ jobs:
     steps:
       - uses: actions/checkout@v3
 
-      - uses: google-github-actions/setup-gcloud@v0
+      - id: "google-cloud-auth"
+        name: "Authenticate to Google Cloud"
+        uses: "google-github-actions/auth@v1"
         with:
-          project_id: cpg-common
-          service_account_key: ${{ secrets.GCP_SERVER_DEPLOY_KEY }}
+          workload_identity_provider: "projects/1051897107465/locations/global/workloadIdentityPools/github-pool/providers/github-provider"
+          service_account: "gh-images-deployer@cpg-common.iam.gserviceaccount.com"
+
+      - id: "google-cloud-sdk-setup"
+        name: "Set up Cloud SDK"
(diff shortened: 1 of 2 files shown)


jmarshall:
Mea culpa, I should have noticed that when I reviewed #117 as we had already run into the same thing in the metamist repo.

I previously made a similar change to this one to fix the same accident in populationgenomics/metamist#550, but we took the opportunity to move to use a federated identity instead in populationgenomics/metamist#553. Possibly that's something we would want to do in this repo too?

illusional:
Hey @EddieLF, @jmarshall beat me to the suggestion, it's unclear to me at this stage which permissions one needs to do this so I've done it and added a couple of commits to your branch to address this.

jmarshall:
This quiet change in action parameters is exactly the sort of thing that actionlint should be able to detect. It knows about a bunch of common actions but so far not this one. However I've built a version locally that does, so hopefully we won't make this mistake again! And also proposed adding it upstream — see rhysd/actionlint#380.

illusional (approved):
Feels weird approving, but good to get this out

illusional:
I'm going to revoke the gh-images-deployer credentials, keep an eye out for anything weird.

PR #119, opened 2023-11-17
Update samtools to 1.18 and include Python dill package

jmarshall:
The upcoming CRAM reheadering script (see populationgenomics/production-pipelines#526) would like to use Hail Python jobs (via `new_python_job`) that shell out to `samtools`. In order to do this, we need a samtools image that has the `dill` Python package on it, as this is [required in order to run Hail Python jobs](https://hail.is/docs/batch/api/batch/hailtop.batch.job.PythonJob.html#hailtop.batch.job.PythonJob.image).

Update to the latest samtools just for good measure (and to keep this separate from the existing image already in widespread use).


illusional (approved):

jmarshall:
Thanks Michael. My finger slipped over the assignment button by accident, but your review is also appreciated 😄

jmarshall:
Deployment failed because now I'm having @milo-hyben's micromamba downloading problem. Oops!

PR #120, opened 2023-11-19
Use a similar download URL to those micromamba itself uses

jmarshall:
The initial download of micromamba*.tar.bz2 from api.anaconda.org has recently begun failing when run under docker, presumably due to network setup changes elsewhere.

See https://centrepopgen.slack.com/archives/C03FZL2EF24/p1699837108996249 for context.

Fortunately `micromamba install …` itself still works from docker, so we can use `micromamba install -v` to observe the URL style it uses when downloading and installing other packages, and use the same hostname and URL style to do our initial manual download of the micromamba package.


illusional (approved):
Build worked!

PR #121, opened 2023-11-20
Ensure dill etc are installed for Python 3.10

jmarshall:
Final fix to make a samtools image that Hail can use for its Python jobs: hail needs to run dill using the same Python version as the driver image.

This image has been built into …/images-dev and successfully used for a test run of CRAM reheadering: see <https://batch.hail.populationgenomics.org.au/batches/430054>.


illusional (approved):
What's the difference between micromamba install dill and a pip install? If it works I'm happy, just curious.

jmarshall:
I was somewhat expecting that image not to have a `pip` command installed, but it does. And rather to my surprise, `pip list` reports dill as having been installed by pip — probably because the conda-forge build script did a `pip install` in the process of making the package tarball.

So in the end there's really no difference here (at least for a noarch python package). The files installed are the same, they've just come from unpacking a conda package rather than installing an sdist or wheel from pypi.

PR #131, opened 2024-02-19
Create Dockerfile of TRTools to include the --vcf-list option

hopedisastro:
This is a temporary image (to get us access to the `--vcf-list` option) that will get deleted once TRTools comes out with a tagged release, which we will use to then correspondingly update our `trtools` image version.


MattWellie:
If you want this pushed & available via config you'll also need the image name/version added here: https://github.com/populationgenomics/images/blob/main/images.toml

hopedisastro:
closing because the v6 release came out! so will update the trtools image version instaed

PR #132, opened 2024-02-28
Minimal gcsfuse image

EddieLF:
Installs only gcsfuse, for mounting buckets in Hail jobs with write privileges.

### DIFF
--- a/images/gcsfuse/Dockerfile
+++ b/images/gcsfuse/Dockerfile
@@ -1,0 +1,8 @@
+FROM ubuntu:22.04
+
+ARG VERSION=${VERSION:-1.0}
+
+RUN GCSFUSE_REPO=gcsfuse-$(lsb_release -c -s) && \
+    echo "deb http://packages.cloud.google.com/apt $GCSFUSE_REPO main" | tee /etc/apt/sources.list.d/gcsfuse.list && \
+    wget -qO - https://packages.cloud.google.com/apt/doc/apt-key.gpg | apt-key add - && \
+    apt-get update && apt-get install -y gcsfuse


images/gcsfuse/Dockerfile
```diff
@@ -0,0 +1,8 @@
+FROM ubuntu:22.04
+
+ARG VERSION=${VERSION:-1.0}
+
+RUN GCSFUSE_REPO=gcsfuse-$(lsb_release -c -s) && \
+    echo "deb http://packages.cloud.google.com/apt $GCSFUSE_REPO main" | tee /etc/apt/sources.list.d/gcsfuse.list && \
```
jmarshall, line 6: This needs to be https; see [install docs](https://cloud.google.com/storage/docs/gcsfuse-install#install-binaries) and GoogleCloudPlatform/gcsfuse#1424.

```suggestion
    echo "deb https://packages.cloud.google.com/apt $GCSFUSE_REPO main" | tee /etc/apt/sources.list.d/gcsfuse.list && \
```

EddieLF:
Excellent catch, thank you @jmarshall

jmarshall:
The `https` fix fell off in the merge commit, and I guess in due course this will need to be added to _images.toml_.

images/gcsfuse/Dockerfile
```diff
@@ -0,0 +1,10 @@
+FROM ubuntu:22.04
+
+ARG VERSION=${VERSION:-1.0}
+
+RUN apt-get update && \
+    apt-get install -y lsb-release wget && \
+    GCSFUSE_REPO=gcsfuse-$(lsb_release -c -s) && \
+    echo "deb http://packages.cloud.google.com/apt $GCSFUSE_REPO main" | tee /etc/apt/sources.list.d/gcsfuse.list && \
```
MattWellie, line 8:
```suggestion
    apt-get install -y gnupg lsb-release wget && \
    GCSFUSE_REPO=gcsfuse-$(lsb_release -c -s) && \
    echo "deb https://packages.cloud.google.com/apt $GCSFUSE_REPO main" | tee /etc/apt/sources.list.d/gcsfuse.list && \
```

I tried this locally and it wouldn't build without gnupg/similar installed

```E: gnupg, gnupg2 and gnupg1 do not seem to be installed, but one of them is required for this operation```

illusional:
Dismissing my review because you've got some good engaged feedback here

PR #133, opened 2024-03-06
Image to run Storey's q-value R package

hopedisastro:
Image to run the `qvalue` package in R. Also installs a python wrapper to run in R. 
Script that uses this image:  https://github.com/populationgenomics/sv-workflows/pull/165

I have tested this with the script in `test` and works nicely.


MattWellie (approved):
Based on the script you linked there's no need to pip install analysis-runner, cpg-workflows, or gcsfuse.

PR #135, opened 2024-03-24
Exomiser

MattWellie:
Creates an image for Exomiser, based on Giles' implementation [here](https://github.com/broadinstitute/be-s-t/tree/gh/gcs-exomiser/scripts/exomiser)

This is not the latest exomiser vesion, but... it works 🙃

I tried installing the latest and greatest, but attempts to use 14.0.0 failed miserably


cassimons (approved):
Looks good. FWIIW With that set of filters REMM can't actually do much can it?

MattWellie:
> Looks good. FWIIW With that set of filters REMM can't actually do much can it? 

Probably not, we can revise the settings but I'm conscious of Giles' cost explosion when he relaxed all the default parameters

cassimons:
Totaly fair. My thought was more we could look at dropping remm as i am suss on the value it would add in any case. All good as it is though.

PR #138, opened 2024-04-17
Environment Lock

nevoodoo:
## Description
In a bid to introduce better security permissions, we have replaced the use of SA keys to workload identity providers where possible. We will also ensure the workload federated identities are attached to an environment in the repo. 

## Changes
- [x]  Updated the workflow to use WIP and separated it into two jobs, one for prod and one for dev
- [x]  Added `environment` attribute to the `deploy_*` workflows


MattWellie (approved):
TLGM

PR #142, opened 2024-04-29
Exomiser 14

MattWellie:
Adds a separate build for Exomiser 14.X so as not to conflict with the current 13.X version (the two versions do not share db files, so these are incompatible)

Reliant on the data being sourced separately (see https://github.com/populationgenomics/references/pull/55)


EddieLF (approved):
Approving assuming that the linting can be fixed. I don't really understand the error mypy is giving, is it related to this issue? https://github.com/python/mypy/issues/10768

MattWellie:
> I don't really understand the error mypy is giving

AFAIK because I've bundled the same python script into both exomiser builds, and neither folder was set up as a python module, mypy detects two files with the same name in the same namespace and cries. TBH this is a mostly non-python repo, so mypy as a pre-commit hook isn't very useful. I've solved/mitigated by adding an init.py, so each folder reads as a separate python module

PR #143, opened 2024-05-02
Fix VEP_105 builder

### LINKED ISSUE #111: VEP failed to build...
Current config entry:
```
[images]
vep_105 = "australia-southeast1-docker.pkg.dev/cpg-common/images/vep_105:105.0"
```

But this image failed to generate: [here](https://github.com/populationgenomics/images/actions/runs/6568720187). The version of VEP 105 that exists is:

```
[images]
vep = "australia-southeast1-docker.pkg.dev/cpg-common/images/vep:105.0"
```

Which now doesn't appear in the config. Either the VEP 105 image definition needs to be fixed, or the working one needs to be Skopeo'd to the location appearing in config.

I've treated this issue by overriding the config value for `vep_105` in the seqr_loader default config, but we should rebuild this eventually

MattWellie:
See https://github.com/populationgenomics/images/pull/132#discussion_r1506728301

Closes #111

### DIFF
--- a/images/vep_105/Dockerfile
+++ b/images/vep_105/Dockerfile
@@ -46,7 +46,7 @@ RUN apt-get update && \
     rm -rf loftee_38 && \
     # Install gcsfuse to mount VEP cache files
     GCSFUSE_REPO=gcsfuse-$(lsb_release -c -s) && \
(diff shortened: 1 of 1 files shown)


EddieLF (approved):
Looks secure to me

PR #146, opened 2024-05-23
Create new image for openpyxl and pandas

amyminiter:
Amy needs openpyxl to run a script via `analysis-runner`. This script creates a Docker file that appends to the standard `analysis-runner` image.


.gitignore
```diff
@@ -1,2 +1,4 @@
 .DS_Store
 .idea
+
+__amy_scratch
```
jmarshall, line 4: You probably didn't mean to commit this…

There's a local ignore file in _.git/info/exclude_ in your checked-out repo directory — if you add `__amy_scratch` to that, it'll be ignored just for you and there's no risk of checking it in.

(Also use _~/.config/git/ignore_ for locally ignoring things in all the repos on your machine.)

michael-harper (approved):

.gitignore
amyminiter: Thanks John. Good tip

PR #149, opened 2024-06-10
Add susieR to the dockerfile

hopedisastro:
I am making `r-meta` as an image for meta-analysis and downstream analysis after eQTL discovery - it will have `coloc` and `susieR` packages for colocalisation and finemapping analysis respectively. 

Confirming the image works in my test scripts.


michael-harper (approved):
If it's working in your testing, I can't see any reason why it wouldn't work on main

PR #160, opened 2024-08-21
Add Aspera image

illusional:
Migrated from the [hgdp repository](https://github.com/populationgenomics/hgdp/tree/main/download_crams), and upgraded to a later version of Aspera.

We want to use the root user, AND we want gcloud, which rules out the official container (martinlaurent/ascli:4.18.0)


illusional:
Not sure if you're using this @MattWellie, but I'll close my copy of it :)

MattWellie:
Cheers @illusional, I'm not sure of the use-case for this. I've notified the ourdna folks in case this version of the client is still required, and I'll hold off cleaning up the branch just in case 👌

MattWellie:
Happy to nuke this for now, if we do need the aspera client we'll recreate the aspera + gcloud install, but at the moment this isn't sounding super likely

PR #162, opened 2024-09-17
latest saige-qtl 

annacuomo:
This seems to be the latest commit? 

https://github.com/weizhou0/qtl/commit/aaaadeba8748aecebc1268252567216a1272c59a


jmarshall (approved):
Yep, I concur that's the most recent commit.

PR #170, opened 2024-10-21
"AIP" to talos

MattWellie:
Use the renamed repository when the Talos image is built

And a version bump to boot


MattWellie:
There's been about 6 new versions since this was opened 😓

cassimons (approved):

PR #171, opened 2024-10-31
Update TRTools to v6.0.2

hopedisastro:
This version incorporates the --vcfs-list parameter that will hopefully resolve the mergeSTR issuehttps://centrepopgen.slack.com/archives/C030X7WGFCL/p1730386942089209. 

Had to change ARG VERSION to 'VER' because it was automatically adding an extra 'v' to 6.0.2 - resulting in a failure to find the version in conda


jmarshall:
When you're manually running the `deploy_container` workflow, are you putting `v6.0.2` for _Tag of image_?

I think that's where the mystery `v` is coming from. Type in `6.0.2` instead — which is what the automatic run will do when this is eventually merged — and I think you will be able to remove all this hardcoding. :smile:

images/trtools/Dockerfile
```diff
@@ -2,13 +2,14 @@ FROM debian:buster-slim
 
 ENV MAMBA_ROOT_PREFIX /root/micromamba
 ENV PATH $MAMBA_ROOT_PREFIX/bin:$PATH
-ARG VERSION=${VERSION:-5.0.2}
+ARG VER=6.0.2
```
hopedisastro, line 5:
```suggestion
ARG VERSION=${VERSION:-6.0.2}
```

images/trtools/Dockerfile
```diff
@@ -2,13 +2,14 @@ FROM debian:buster-slim
...
+RUN apt-get update && apt-get install -y git curl bash bzip2 zip && \
     rm -r /var/lib/apt/lists/* && \
-    rm -r /var/cache/apt/* && \
-    wget -qO- https://api.anaconda.org/download/conda-forge/micromamba/0.8.2/linux-64/micromamba-0.8.2-he9b6cbd_0.tar.bz2 | tar -xvj -C /usr/local bin/micromamba && \
+    rm -r /var/cache/apt/*
+RUN curl -Ls https://micro.mamba.pm/api/micromamba/linux-64/latest | tar -xvj bin/micromamba && \
     mkdir ${MAMBA_ROOT_PREFIX} && \
     micromamba install -y --prefix ${MAMBA_ROOT_PREFIX} -c bioconda -c conda-forge \
-    trtools=${VERSION} && \
+    trtools=${VER} && \
```
hopedisastro, line 14:
```suggestion
    trtools=${VERSION} && \
```

hopedisastro:
Ok, works now without hardcoding.

jmarshall (approved):
I think with that `v` sorted out, the micromamba update wouldn't be necessary — but it's an improvement anyway so let's leave it in!

PR #175, opened 2024-12-11
Illumina connected analytics image

michael-harper:
PA team is implementing ICA to realign and genotype datasets. This image contains the ICA CLI that will facilitate downloading data from the ICA platform to GCP


Alexander-Stuckey (approved):
Looks good to me

PR #178, opened 2024-12-19
base image off cpg_workflows and bump version

Alexander-Stuckey:
Updated ICA image here, Left `jq` in because I'm not sure if it's in `cpg_workflows` or not.


jmarshall (approved):

jmarshall:
```
6 2.270 jq is already the newest version (1.6-2.1).
#6 2.270 0 upgraded, 0 newly installed, 0 to remove and 94 not upgraded.
[…]
ERROR: failed to solve: process "[…]" did not complete successfully: exit code: 8
```

It appears that `jq` was indeed already present, and `apt` returns an error code if you try to do a no-op install… 🫠

Alexander-Stuckey:
Yeah there was an issue with the version, since it was using that arg to download the ICA CLI tool and not just to tag the resultant container. So have to keep it static.

PR #181, opened 2025-01-20
 Install pip inside ICA image

michael-harper:
Hitting issues with running ICA pipeline using standalone ICA image [here](https://batch.hail.populationgenomics.org.au/batches/589302/jobs/1)
Adding Dill package to ICA image so that PythonJob arguments can get de/serialised.


jmarshall (approved):

MattWellie (approved):
FWIW I think this can/should be a single  pip install command for the multiple target packages. That way Pip can get ahead of any dependency conflicts

PR #184, opened 2025-02-05
Removing unnecessary step to upload new version of cpg-workflow

milo-hyben:
This PR removes unnecessary step in CI to upload new version of cpg-workflow each time image repo is updated.

It seems whenever an update to image repo is merged into main, CI will rebuild new cpg-workflow docker image and push it to the Artifact registry, and it really needs only the cpg-workflow version for the configuration.
Version gets picked in the beginning of the step 'deploy_cpg_workflows'.

cpg-workflow has its own CI to deploy its docker image whenever changes are merged in the repo. 

Note: non tagged version of cpg-workflow in production Artifact Registry are created by removed CI section in this PR.


MattWellie:
There's a reason for this workflow, even if it's not a good reason - this repository builds the `images` section of the global config template used in analysis-runner.

New merges to main in production-pipelines pushes the latest version of that codebase and image using the workflow [here](https://github.com/populationgenomics/production-pipelines/blob/main/.github/workflows/docker.yaml), but they don't update the images template. Until this workflow runs, the image template will contain a `cpg_workflows` entry that points to the latest version of the image pushed by this workflow.

https://github.com/search?q=repo%3Apopulationgenomics%2Fproduction-pipelines+%22image_path%28%27cpg_w%22&type=code

milo-hyben:
We had discussion regarding at software standup and this PR does not remove the section where config.toml gets updated with the latest version of cpg_workflows image.
It only removes building of the new image. 
I had renamed step deploy_cpg_workflows to reflect the new change.

What is missing and needs to be addressed (not in this PR) is something which has been missing for a while, updating config.toml whenever there is a new version of production-pipelines/cpg_workflows.
I have a suggestion to maybe create a new GitHub action in tis repo to run on a schedule, maybe daily?
This is going to be temporary anyway as we are moving to CPG Flow, where file config.toml is going to be redundant.

jmarshall (approved):

PR #185, opened 2025-02-10
Show link to built image as last step of the CI.

milo-hyben:
This is implementation for task: 
https://cpg-populationanalysis.atlassian.net/browse/SET-414?atlOrigin=eyJpIjoiODViMGFlNWExNjY5NDIzYWJkOGZhMDEzYTgwYWU1MDkiLCJwIjoiaiJ9

After images gets pushed to registry it shows the link in the log output.

### DIFF
--- a/.github/workflows/deploy_container.yaml
+++ b/.github/workflows/deploy_container.yaml
@@ -62,9 +62,11 @@ jobs:
         run: |
           docker tag $DOCKER_IMAGE $DOCKER_MAIN$DOCKER_IMAGE
           docker push $DOCKER_MAIN$DOCKER_IMAGE
+          echo "Link to image: https://console.cloud.google.com/artifacts/docker/cpg-common/australia-southeast1/images/$IMAGE_NAME?project=cpg-common"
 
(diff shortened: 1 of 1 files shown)


jmarshall:
Nice. You can also [echo it `>> $GITHUB_STEP_SUMMARY`](https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/workflow-commands-for-github-actions#adding-a-job-summary) and it will appear on the summary page something like [how these pip-audit notes do](https://github.com/populationgenomics/metamist/actions/runs/13150201511).

milo-hyben:
> Nice. You can also [echo it `>> $GITHUB_STEP_SUMMARY`](https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/workflow-commands-for-github-actions#adding-a-job-summary) and it will appear on the summary page something like [how these pip-audit notes do](https://github.com/populationgenomics/metamist/actions/runs/13150201511).

Amazing @jmarshall, will include it, was not aware of this one. Thank you!

jmarshall (approved):

PR #191, opened 2025-02-26
SET-416-convert-existing-images

nevoodoo:
## Description
https://cpg-populationanalysis.atlassian.net/browse/SET-416?atlOrigin=eyJpIjoiMGI0MTk1NGRiMzM0NGMzYTk1NDg4MzlkYzk0Yzc0M2YiLCJwIjoiaiJ9
As an analyst,
I want a defined versioning and tagging policy for container images,
so that we can revise the contents of an existing container and use that new container while maintaining the ability to reproduce analyses running on that previous container.

## Proposal
In this PR, the goal is to add one main functionality:
- [x] When a PR is merged into `main` and it contains changes to any image Dockerfiles, a github action will recognise all the changed images, and fetch the next available tag version and build/push the images to the container registry. - **✅ Actioned in `revamp.yaml`** ✅ (will eventually replace current docker workflow)

### To be done in a separate PR
- [ ] Add the ability to automatically build dev images and deploy them to `images-dev`.

## Notes
- As per discussions, the  use of `images.toml` file to maintain version numbers can probably be deprecated in favour of using `ARG` or `ENV` within the Dockerfiles.


.github/workflows/get_expected_tags.yaml
```diff
@@ -73,37 +73,19 @@ jobs:
...
-          fi
-
-      - name: Update or Create PR Comment
-        env:
-          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
-        run: |
-          COMMENT_BODY="### 🐳 Dockerfile Version Changes\n\n$TABLE"
-          COMMENTS_URL="https://api.github.com/repos/${{ github.repository }}/issues/${{ github.event.pull_request.number }}/comments"
+      - name: Find Comment
+        uses: peter-evans/find-comment@v3
```
jmarshall, line 78: I'd suggest using [`gh pr comment`](https://cli.github.com/manual/gh_pr_comment) (already installed on ubuntu-latest) rather than a random third-party action.

nevoodoo: I've vetted the action, I personally think it is safe to use - created by staff member at Github and has good support. Happy to reconsider if we still feel strongly about it ;)

violetbrina (commented):
Looking good on paper. Keen to see if the test works. How is this being enforced to only run on the dummy image is my only question.

.github/workflows/get_version.py
```diff
@@ -0,0 +1,134 @@
...
+import json
+import subprocess
+import re
+import os
+
+
+def extract_version_from_file(file_path):
+    """
+    Extract the version from a Dockerfile by searching for a line like:
+      ENV VERSION=1.0.0
```
MattWellie, line 11: `ARG VERSION=${VERSION:-XXXX}` is a common versioning pattern, so that we can build new versions from unchanged dockerfiles

e.g. https://github.com/populationgenomics/images/blob/main/images/samtools/Dockerfile#L5

nevoodoo: Yeah, this is something I plan to do once I build out the workflow_dispatch functionality. Ideally we'd want to be able to build new versions from unchanged dockerfiles, while maintaining the convention. Is there a use case where you'd want to tag it by something that isn't the standard convention? If not, we might be able to do away with the need to have the `ARG`

MattWellie: I was just mulling this over, and I don't know why we have `ARG VERSION=${VERSION:-XXXX}`... This just uses the Bash substitution "if VERSION is set, use it, otherwise take the default value XXXX"

That's the same behaviour as `ARG VERSION=XXXX`... which would be much easier to detect with RegEx.

nevoodoo: Haha, yes that was exactly my thought process, except I used ENV instead. I think the ARG is useful for the workflow_dispatch and the substitution may have been to override the tag to a custom version

MattWellie: I guess the main thing I wanted to flag here is that the versions inside the Docker image aren't always the latest in-use version. Talos as an example:
- the build [file](https://github.com/populationgenomics/images/blob/main/images/talos/Dockerfile#L3) has version 6.4.0
- the [images.toml](https://github.com/populationgenomics/images/blob/main/images.toml#L50) is consistent with that, but if it was different in the TOML file that would override the version in the Dockerfile when the workflow runs
- I've found when software is regularly updated, requiring an approved PR to deploy the new version can be onerous, so I've been using the [CI workflow](https://github.com/populationgenomics/images/actions/workflows/deploy_container.yaml) directly, and the latest build from main is actually `6.6.3`

All that is to say that the regular expression on the docker file would find `a` version, but that's not a guarantee it's the latest version in the wild, unless we get rid of the toml file as a build input, and remove the individual image builder, forcing the Dockerfile to be a single source of truth

nevoodoo: I think removing the need to maintain the toml file was the intention here as you rightly pointed out. It doesn't seem to make too much sense to be maintaining versions in multiple places.

nevoodoo:
https://github.com/populationgenomics/images/actions/runs/13957446472?pr=191

This is a successful test run where we tested an on push event to this branch and it successfully tagged and pushed the image.

.github/workflows/revamp.yaml
```diff
@@ -0,0 +1,194 @@
...
+
+      - id: 'google-cloud-sdk-setup'
+        name: 'Set up Cloud SDK'
+        uses: google-github-actions/setup-gcloud@v2
+        with:
+          project_id: ${{ env.PROJECT }}
+
+      - name: 'Get current version'
+        id: get_next_version
+        run: |
```
ckrishk, line 49: What happens if this step fails ? Does the workflow or the script handle it ?

.github/workflows/revamp.yaml
```diff
@@ -0,0 +1,194 @@
...
+          next_version=$(python .github/workflows/get_version.py)
+          echo "next_version=$next_version" >> "$GITHUB_OUTPUT"
+
+  # Deploy the images that have changed to the registry
+  deploy_images:
+    runs-on: ubuntu-latest
```
ckrishk, line 55: What does Ubuntu-latest refer to in this case? Is this the image that is tested for vulns . Do we want to consider pinning this to a suitable version and manage upgrades as needed ?

.github/workflows/revamp.yaml
```diff
@@ -0,0 +1,194 @@
...
+    environment: production
+    needs:
+      - get_next_version
+    if: ${{ needs.get_next_version.outputs.matrix != '{}' && needs.get_next_version.outputs.matrix != '' }}
+    strategy:
+      matrix: ${{ fromJson(needs.get_next_version.outputs.matrix) }}
+    env:
+      DOCKER_BUILDKIT: 1
+      BUILDKIT_PROGRESS: plain
+      CLOUDSDK_CORE_DISABLE_PROMPTS: 1
```
ckrishk, line 65: Nice!

.github/workflows/revamp.yaml
```diff
@@ -0,0 +1,194 @@
...
+    steps:
+      - uses: actions/checkout@v4
+
+      - id: 'google-cloud-auth'
+        name: 'Authenticate to Google Cloud'
+        uses: 'google-github-actions/auth@v2'
+        with:
+          workload_identity_provider: 'projects/1051897107465/locations/global/workloadIdentityPools/github-pool/providers/github-provider'
+          service_account: 'gh-images-deployer@cpg-common.iam.gserviceaccount.com'
```
ckrishk, line 74: Is this open source repo? My 2 cents, turn this into a GitHub workflow variable, username and secrets should not be exposed in plain text

.github/workflows/revamp.yaml
```diff
@@ -0,0 +1,194 @@
...
+          cd production-pipelines
+          echo "sha=$(git rev-parse HEAD)" >> $GITHUB_OUTPUT
+          cd -
+
+      - id: 'google-cloud-auth'
+        name: 'Authenticate to Google Cloud'
+        uses: 'google-github-actions/auth@v2'
+        with:
+          workload_identity_provider: 'projects/1051897107465/locations/global/workloadIdentityPools/github-pool/providers/github-provider'
+          service_account: 'gh-images-deployer@cpg-common.iam.gserviceaccount.com'
```
ckrishk, line 127: Same as above

.github/workflows/revamp.yaml
```diff
@@ -0,0 +1,194 @@
...
+          IMAGE_NAME=${{ env.IMAGES_PREFIX }}/cpg_workflows
+          docker push $IMAGE_NAME:${{ steps.get_sha.outputs.sha }}
+
+      - name: push latest
+        run: |
+          IMAGE_NAME=${{ env.IMAGES_PREFIX }}/cpg_workflows
+          docker tag $IMAGE_NAME:${{ steps.get_sha.outputs.sha }} $IMAGE_NAME:${{ steps.get_version.outputs.version }}
+          docker tag $IMAGE_NAME:${{ steps.get_sha.outputs.sha }} $IMAGE_NAME:latest
+          docker push $IMAGE_NAME:${{ steps.get_version.outputs.version }}
+          docker push $IMAGE_NAME:latest
```
ckrishk, line 152: Did we agree we will not use latest tag ?

ckrishk (commented):
Looking good 👍 thanks for putting this together Yash. Have a few comments and open questions which should take us close to merge . Thanks 🙏

.github/workflows/revamp.yaml
nevoodoo: Good point! I've removed the latest tags here which were only for `cpg_workflows` anyway

.github/workflows/revamp.yaml
nevoodoo: Yep, it will just skip if there is no output from that step (usually meaning no Dockerfiles have changed)

.github/workflows/revamp.yaml
nevoodoo: Believe there was an issue recently with pinning this and we agreed to just use `ubuntu-latest` instead.

.github/workflows/revamp.yaml
nevoodoo: With how OIDC works, this shouldn't be a security risk as it is not a sensitive secret. We have this publicly in all our repos but I suppose moving them to a secret wouldn't hurt either. Perhaps something to be done as part of a larger cleanup since I'd have to make the change in multiple repos.

.github/workflows/revamp.yaml
jmarshall: It's the [standard good Linux actions runner](https://docs.github.com/en/actions/using-github-hosted-runners/using-github-hosted-runners/about-github-hosted-runners#standard-github-hosted-runners-for-public-repositories). There are some arguments for using `ubuntu-24.04` instead, but they're just trade-offs; generally IMHO `-latest` is the right answer and it is what we use in all our GitHub Actions workflows throughout CPG's repositories.

ckrishk (approved):

jmarshall (approved):

PR #193, opened 2025-03-13
Set 465 implement & document out image management plan for data managers

violetbrina:
# Why?
To enable data managers to archive and unarchive images independent of software

# How?
Implementing a github actions workflow that will read an `archive.yaml` and `unarchive.yaml` file respectively which on approval and merge to main will run and update the Artifact Registry state appropriately (similar to how cpg-infra-private permission PRs work).


violetbrina:
## Notes:
First commits - Just moving code over.

Currently my local pre-commit check of mypy doesn't work and is commented out. I'd be curious if this is a me issue or not.

Didn't have time for much of a refactor yet. Mostly just a file dump.

vivbak:
@jmarshall :)

violetbrina:
Can't leave a GitHub review haha but I've had a look through @dancoates. Looks good to me. Keen to see a test run of it as well if that exists.

violetbrina:
Actually is the code working on this branch? I'd be curious to know if doing an archive then unarchive of a single image would work or not.

dancoates:
> Actually is the code working on this branch? I'd be curious to know if doing an archive then unarchive of a single image would work or not.

Yep it should be, I tested by unarchiving and then rearchiving an old cpg_workflows image. You can run the script manually if you'd like to test it out.

scripts/common/image_repository_helpers.py
```diff
@@ -173,16 +206,139 @@ def from_artifact_repository_image(image_data: DockerImage) -> 'Image':
...
+    images: list[Image]
+
+    def includes_image_version(self, image: Image):
+        """
+        Check if this repository already has a certain version of an image
+        """
+        matching_images = [
+            i for i in self.images if i.name == image.name and i.digest == image.digest
+        ]
+        return len(matching_images) > 0
```
jmarshall, line 220: Bit trivial, but these four lines seem like a job for `any(…)`.

scripts/archive_images.py
```diff
@@ -0,0 +1,136 @@
...
+
+
+def get_archive_list():
+    archive_list_file = Path(__file__).parent.parent / 'archived_images.txt'
+    archive_list: list[str] = []
+    with open(archive_list_file) as f:
+        for line in f:
+            archive_list.append(line.strip())
+    # remove duplicates from the list and return
+    return list(set(archive_list))
```
jmarshall, line 28: This is used in `archive_images_in_repository()` to do a whole bunch of `if x [not] in archive_list` operations. Having this as a list makes `archive_images_in_repository()` O(n<sup>2</sup>).

I'd recommend creating this as a `set[str]` and returning it as a `set[str]`. Apologies if this means renaming the function and variables to `…_set` :smile:

scripts/common/image_repository_helpers.py
dancoates: Good point, have updated

scripts/archive_images.py
dancoates: Thanks for that pickup, for some reason I had in my head that `in` on a list in python was `O(n)` but you're absolutely right, have updated to return and pass around a set.

scripts/archive_images.py
```diff
@@ -0,0 +1,134 @@
...
+)
+
+logging.getLogger().setLevel(logging.INFO)
+
+SUPPORTED_REPOSITORIES = ['images']
+
+
+def get_archive_set():
+    archive_list_file = Path(__file__).parent.parent / 'archived_images.txt'
+    archive_list: list[str] = []
```
jmarshall, line 23: It could be a set here too, and add each line with `archive_set.add(…)`, but no biggie.

scripts/common/image_repository_helpers.py
```diff
@@ -173,16 +206,139 @@ def from_artifact_repository_image(image_data: DockerImage) -> 'Image':
...
+
+    try:
+        # Copy the image from the source repo to the destination. This doesn't need
+        # to check if it already exists as skopeo won't do anything if the image is
+        # already there, and it is pretty much as quick as running an additional check
+        logging.info(f'Copying image from {source} to {destination} using skopeo...')
+
+        subprocess.run(  # noqa: S603 - shlex escapes the command, linter just can't tell
+            shlex.split(
+                f'skopeo copy --all --preserve-digests docker://{source} docker://{destination}'
```
jmarshall, line 263: I would probably have written it explicitly in list form — `subprocess.run(['skopeo', 'copy', …], check=True)` — and not needed any escaping or shlex. But that still gets S603 anyway 🤷 so this is fine too.

jmarshall (approved):
A couple of food for thought comments that don't really need any changes to be made, otherwise LGTM thanks.

jmarshall:
Nice catch on the installing skopeo thing… I guess that means the workflow hasn't been exercised or tested, but I assume the script itself has locally.

dancoates:
> Nice catch on the installing skopeo thing… I guess that means the workflow hasn't been exercised or tested, but I assume the script itself has locally.

Yeah that's right, the workflow needs testing, it's a pretty simple one so hopefully not too many problems, I expect we'll need to tweak permissions though

scripts/archive_images.py
dancoates: yeah good point, may as well avoid the conversion

PR #195, opened 2025-03-17
Echtvar joint gnomad

MattWellie:
Builds an Echtvar config to steal a wide range of fields from the joint gnomAD v4, instead of the previous config which was only for genomes

Floating this here in case we want to adjust the config file, taking even more fields


EddieLF (approved):

MattWellie:
Fixing this config here, given that we will only use this data for Talos, so I'm happy with the content

PR #196, opened 2025-03-18
SET-409-Improve-Automated-Tagging-Visibility

nevoodoo:
References: https://cpg-populationanalysis.atlassian.net/browse/SET-409

And is broken out from #191 to improve PR readability.

## Proposal
When a PR is created and it contains changes to any image Dockerfiles, a github action will recognise all the changed images, and fetch expected tag versions for when the PR is merged, and add them as a comment to the PR for better visibility.** - ✅ Actioned in `get_expected_tags.yaml` ✅

I've also created a python script to compare commits between HEAD and the latest commit in a feature branch for PR events, as well as the two latest commits in the `main` branch for the `on-push` event. - ✅ Actioned in `get_version.py` ✅

Also pushing a `dummy` image that we can safely use for testing and reference in other workflow tests.

## Notes
- **There is an edge case where multiple PRs amending the same Dockerfiles, will all have similar expected tag versions, but only the first to be merged will receive said expected tag version.


.github/workflows/get_expected_tags.yaml
```diff
@@ -0,0 +1,91 @@
...
+  id-token: write
+  contents: read
+
+env:
+  PROJECT: 'cpg-common'
+
+jobs:
+  get_next_version:
+    runs-on: ubuntu-latest
+    environment: production
```
jmarshall, line 17: This `environment`, the `[env] PROJECT` above, and the two google-cloud steps can all be removed I think.

jmarshall:
See PR #198 for how the `find-comment` and `create-or-update-comment` actions can be replaced by a `gh pr comment` one-liner. IMHO that would be simpler.

That needs a repo checkout, and the best way to do that would be to refactor this workflow so it runs in one job instead of two, which I think would be an improvement for this single-purpose split-out workflow anyway.

I believe you could also then remove the `permissions` stanzas, as `gh` requires only default permissions to do this.

.github/workflows/get_expected_tags.yaml
nevoodoo: I tested this earlier to check if we'd need the `gcloud` auth stuff and can confirm these steps are necessary to fetch the tags from the artifact registry, without which this sticks the default suffix of `-1` which isn't right.

.github/workflows/get_version.py
```diff
@@ -0,0 +1,134 @@
...
+        cmd = [
+            'gcloud',
+            'container',
+            'images',
+            'list-tags',
+            full_image_name,
+            '--format=json',
+        ]
+        result = subprocess.run(cmd, capture_output=True, text=True)
+        if result.returncode != 0:
```
jmarshall, line 50: I finally figured out why my branch was “working” despite having no GCP credentials: the `gcloud` command was returning

```
result.returncode=1
result.stderr='ERROR: (gcloud.container.images.list-tags) You do not currently have an active account selected. […]'
```

and it was just returning via this “nothing found” `return` here. What is this `if` intended to handle, and can we make it more specific so this script fails in most circumstances when this command fails?

In my testing, GCP returns `[]` successfully when there are no such tagged images, e.g.,

```
gcloud container images list-tags australia-southeast1-docker.pkg.dev/cpg-common/images/talosbork --format=json
```

So this `if` seems unnecessary if it's intended to cater for that case.

ckrishk (approved):
LGTM

.github/workflows/get_version.py
nevoodoo: Excellent points @jmarshall. I've fixed that now, please review to see if I've applied a better approach to this now!

jmarshall:
This is looking good and ready to go; I have two remaining issues:

* OMG can we disable `S603` and `S607` globally or at least globally for this script? What a waste of time.
* When an exception is caught, it should be because it is thrown in the normal operation of the script and is in some way expected. If the exception indicates that something has systematically broken down, we want to see it so we can fix it. So when we get together we should go over each of the `except` clauses and identify what the circumstances for that exception being thrown are.

PR #198, opened 2025-03-18
Update SET-409 PR to use GitHub CLI to add/update PR comments 

jmarshall:
This is a GitHub product that is already installed on the runner image, and should be more immune to supply chain attacks than these (already fairly safe) GH actions.


jmarshall:
The bot's comment was edited, but it stayed at `1.1-1` presumably because the other action hasn't run and there still isn't a `dummy-1.1-1` image in the repository…

jmarshall:
Exploration PR has served its purpose, and the main PR now uses `gh comment`.

PR #203, opened 2025-03-24
Fixing digest extraction issue from docker push command.

milo-hyben:
This PR should extract the digest of pushed images correctly.


.github/workflows/deploy_container.yaml
```diff
@@ -61,14 +61,14 @@ jobs:
         if: ${{ github.ref_name == 'main' }}
         run: |
           docker tag $DOCKER_IMAGE $DOCKER_MAIN$DOCKER_IMAGE
-          digest=$(docker push $DOCKER_MAIN$DOCKER_IMAGE | tee /dev/tty | grep "digest:" | awk '{print $2}')
```
jmarshall, line 64: I guess the `tee` didn't work because the runner uses a non-interactive shell so there is no _/dev/tty_. It would be good to still see the `docker push` output in the logs, which could be arranged with

```
docker push $DOCKER_MAIN$DOCKER_IMAGE | tee push.log
digest=$(grep "digest:" push.log | grep -o 'sha256:[0-9a-f]*')
```

Is `$2` changing to `$3` because the format changed, or because it was just always wrong? Either way it might be good to pick out the sha256 explicitly, e.g. with the `grep -o` above, so it works regardless of the exact output format.

milo-hyben: Excellent suggestion John, that $2 vs $3 was my mistake.
I had tested the latest in my test repo so the workflow should work now.

jmarshall (approved):

jmarshall:
@milo-hyben: Feel free to merge this.

PR #208, opened 2025-03-27
update the ica image with the latest code changes

Alexander-Stuckey:


MattWellie (approved):
Just an FYI - if you want to pump out a new version but the Dockerfile isn't changed, you can use [this](https://github.com/populationgenomics/images/actions/workflows/deploy_container.yaml) workflow, with the `Tag of Image` set to the VERSION you need (here 2.32.7)

Screenshot 2025-03-27 at 11 49 42 AM

PR #213, opened 2025-04-11
Add a new dev deploy workflow

nevoodoo:
The current workflow only supports pushing manually with workflow_dispatch and also allows pushing to the prod registry which is unsafe. The new workflow adds automatic pushes to the dev registry, and changes the behaviour of the workflow_dispatch event to push to the dev registry only.


.github/workflows/revamp_dev.yaml
```diff
@@ -0,0 +1,178 @@
...
+      - name: 'gcloud docker auth'
+        run: |
+          gcloud auth configure-docker ${{ env.DOCKER_PREFIX }}
+
+      - name: Validate and export inputs
+        shell: bash
+        run: |
+          IMAGE_NAME="${{ github.event.inputs.image_name }}"
+          IMAGE_TAG="${{ github.event.inputs.image_tag }}"
+          CLI_ARGS="${{ github.event.inputs.docker_cli_args }}"
```
jmarshall, line 138: If these are going to be carefully validated, we should probably do the `env` thing here:

```
env:
  IMAGE_NAME: …
  IMAGE_TAG: …
  CLI_ARGS: …
```

nevoodoo: Sorry, forgot to push this it seems, done now!

jmarshall (approved):

PR #214, opened 2025-04-11
Archive the kitchen sink

dancoates:
This image is not in use and was created as a test of using VMs for interactive analysis. As such it is safe to archive


vivbak (approved):
I'm exciteeedd!!! 👀 🙌 🚀

PR #218, opened 2025-04-17
Add bcftools v1.21 dockerfile

EddieLF:
A bit ugly to now have three separate bcftools images, however this release comes with htslib 1.21, which has an important fix for writing indexes with the `--write-index` option while not using multiple threads. 

See a Slack discussion of a bad index being written using bcftools 1.20 [here](https://centrepopgen.slack.com/archives/C082NA58R7S/p1744173388834459?thread_ts=1744163974.749869&cid=C082NA58R7S), and the htslib 1.21 release notes detailing the bugfix [here](https://github.com/samtools/htslib/releases/tag/1.21).
> Fix issues with on-the-fly indexing of VCF/BCF (bcftools --write-index) when not using multiple threads.

### DIFF
--- a/images/bcftools_121/Dockerfile
+++ b/images/bcftools_121/Dockerfile
@@ -1,7 +1,8 @@
 FROM debian:bookworm-slim AS base
 
-ARG VERSION=${VERSION:-1.20}
+ARG VERSION=${VERSION:-1.21}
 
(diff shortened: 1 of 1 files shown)


MattWellie (approved):
LGTMatt

PR #220, opened 2025-05-09
Paramiko

EddieLF:
analysis-runner driver image with paramiko python library installed

### DIFF
--- a/images/paramiko/Dockerfile
+++ b/images/paramiko/Dockerfile
@@ -1,4 +1,6 @@
-FROM australia-southeast1-docker.pkg.dev/analysis-runner/images/driver:latest
+ARG DRIVER_TAG="e2aa072191014ba1db0a02eb5e868a4daf327436-hail-d85618aaf4d2bf667aa72c5c6a142e9bafaae9c8"
 
-ARG VERSION=${VERSION:-1.0}
-RUN pip install --no-cache-dir paramiko
+FROM australia-southeast1-docker.pkg.dev/analysis-runner/images/driver:$DRIVER_TAG
+
+ARG VERSION=${VERSION:-3.5.1}
+RUN pip install --no-cache-dir paramiko==${VERSION}
--- a/images/paramiko/Dockerfile
+++ b/images/paramiko/Dockerfile
@@ -1,3 +1,4 @@
 FROM australia-southeast1-docker.pkg.dev/analysis-runner/images/driver:latest
 
+ARG VERSION=${VERSION:-1.0}
 RUN pip install --no-cache-dir paramiko
--- a/images/paramiko/Dockerfile
+++ b/images/paramiko/Dockerfile
@@ -3,4 +3,4 @@ ARG DRIVER_TAG="e2aa072191014ba1db0a02eb5e868a4daf327436-hail-d85618aaf4d2bf667a
 FROM australia-southeast1-docker.pkg.dev/analysis-runner/images/driver:$DRIVER_TAG
 
 ARG VERSION=${VERSION:-3.5.1}
-RUN pip install --no-cache-dir paramiko==${VERSION}
+RUN pip install --no-cache-dir paramiko==3.5.1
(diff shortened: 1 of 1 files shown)


MattWellie (approved):
seems legit. I'd be ok with this building off a specific driver tag instead of latest (to help us streamline not using latest at all)

nevoodoo:
Hey @EddieLF would you mind if we quickly tested a few things with the new images workflows while this PR is open?

EddieLF:
@nevoodoo go ahead, I won't merge til I get the OK

images/paramiko/Dockerfile
```diff
@@ -0,0 +1,4 @@
+FROM australia-southeast1-docker.pkg.dev/analysis-runner/images/driver:latest
+
+ARG VERSION=${VERSION:-1.0}
```
dancoates, line 3: I think in theory this version should reflect the version of paramiko that is being installed? And then be passed on to the pip install command.

nevoodoo: Yep ^ I would put it at 3.5.1 which is the latest version on PyPI. You would only change this version manually if you decided to upgrade to a new version of `paramiko`. Any other changes within this Dockerfile does not need a version bump

EddieLF:
@MattWellie dumb question - how do I find the specific driver tags for the analysis-runner image?

Querying the artifact registry
```
gcloud artifacts docker images list australia-southeast1-docker.pkg.dev/analysis-runner/images/driver --include-tags
```

the latest one is from March 2025, with tag:
```
e2aa072191014ba1db0a02eb5e868a4daf327436-hail-d85618aaf4d2bf667aa72c5c6a142e9bafaae9c8
```

Do I just stick this tag in like:
```
ARG DRIVER_TAG="e2aa072191014ba1db0a02eb5e868a4daf327436-hail-d85618aaf4d2bf667aa72c5c6a142e9bafaae9c8"

FROM australia-southeast1-docker.pkg.dev/analysis-runner/images/driver:$DRIVER_TAG
```
?

Not sure if I want the hail driver image though 🤔

nevoodoo:
> @nevoodoo go ahead, I won't merge til I get the OK
> 
> 

Should be good now, you'll just need to update your branch from main before you merge though!

MattWellie:
> how do I find the specific driver tags for the analysis-runner image?

Yup, nailed it. The analysis-runner driver tags are pretty verbose, but that will work.

Alternatively if you want a lighter image with Hail installed you can use [cpg_hail](https://github.com/populationgenomics/images/blob/main/images/cpg_hail/Dockerfile) as a base.

`australia-southeast1-docker.pkg.dev/cpg-common/images/cpg_hail:0.2.134.cpg1` is the only tag

images/paramiko/Dockerfile
EddieLF: Hmmm so I've set the version to paramiko's version 3.5.1, but when [the deploy is tested](https://github.com/populationgenomics/images/actions/runs/14921294689/job/41917043022?pr=220), it's parsing the `'-1'` alongside the version, so pip is trying to install paramiko==3.5.1-1 - which doesn't exist. 

How do I get around this @nevoodoo ?

nevoodoo: My immediate thought is to just hard code the version number there. Because the dev workflow also supports specifying a tag and that tag gets passed to the `VERSION` arg. So in theory, if you tried to build with a `test` tag, it would try to install `paramiko==test`. This almost sounds like a bug to me tbf. CC @jmarshall 

Instead, maybe have

`ARG VERSION=${VERSION:3.5.1}`
`pip install paramiko=3.5.1`

and then you'd have to bump both these lines when pulling a future release of `paramiko`

MattWellie: Yeah sorry this is definitely a bug. Using the `X` version number to generate a new `X-Y` version is fine, but feeding that arbitrary tag into the Dockerfile as a version is a bug. 

If we're using the dockerfile's contents as a source of truth, we have to leave that unchanged. I'll raise as an issue

PR #223, opened 2025-05-12
Install jq in stripy image

EddieLF:
Adds `jq` to the list of bash modules installed by `apt install`.

This is a useful bash library for json parsing, which is needed for updating the default `config.json` that comes with the stripy source code. See https://github.com/populationgenomics/production-pipelines/pull/1208

NOTE: This PR also hardcodes the stripy version "v2.5". This is because the `$VERSION` now evaluates to `"v2.5-1"` since the images overhaul, and this was preventing successful installation of the module. See comment [here](https://github.com/populationgenomics/images/pull/220#discussion_r2080913770) for an explanation.


MattWellie:
This looks fine, but double-hard coding the version numbers is a result of a bug in the workflow. I've raised https://github.com/populationgenomics/images/pull/225 and #224

MattWellie:
Following the closure of #225 can you drop the duplicated version here? Happy to merge then

PR #225, opened 2025-05-14
Correction to VERSION override

### LINKED ISSUE #224: Library version number overridden in Docker build workflow
Since the image revamp there's a new workflow which uses the VERSION variable in the Dockerfiles + the artifactory state to generate a guaranteed-unique image tag. This image tag is valid as a `tag` for the images it produces.

The newly generated version is also wrongly being into the Dockerfile to override the software Version. This means a VERSION in the Dockerfile is simultaneously being used as a source of truth, and then altered at runtime into a version which probably doesn't exist. 

If the version does exist, the created artefact won't match the VERSION from the Dockerfile.

See thread https://github.com/populationgenomics/images/pull/220#discussion_r2080909960

MattWellie:
Closes #224 

The VERSION in the Dockerfile is being used as a source of truth for the builder workflow when generating a new tag.

The tags are free-form, but the package Version number is not - we can't pass in an alternative VERSION argument at build time, we need the exact version to remain unchanged. See #224

### DIFF
--- a/.github/workflows/revamp.yaml
+++ b/.github/workflows/revamp.yaml
@@ -85,7 +85,6 @@ jobs:
       - name: 'build image'
         run: |
           docker build \
-          --build-arg VERSION=${{ matrix.tag }} \
           --tag ${{ env.IMAGES_PREFIX }}/${{ matrix.name }}:${{ matrix.tag }} \
           images/${{ matrix.name }}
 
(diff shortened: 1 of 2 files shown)


EddieLF (approved):
Yup seems like the fix is to remove it from the build args. Thanks @MattWellie.

PR #226, opened 2025-05-21
Remove legacy deploy_config.yaml and images.toml infrastructure

jmarshall:
These will be replaced by the revamped deployment workflow that does not depend on images.toml.

Also remove the `cpg_workflows` special case from revamp.yaml, as this will be updated directly from _production-pipelines/.github/workflows/docker.yaml_ once populationgenomics/production-pipelines#1219 has been applied.


nevoodoo (approved):
Looks good!

PR #229, opened 2025-05-27
Bump to new hail (gcloud edit.)

MattWellie:
Follow-on from #227 

Uses the new-Hail base image in the hail-gcloud image


images/cpg_hail_gcloud/Dockerfile
```diff
@@ -1,6 +1,6 @@
 FROM australia-southeast1-docker.pkg.dev/cpg-common/images/cpg_hail:0.2.134.cpg2-1
 
-ARG VERSION="0.2.134.cpg2"
```
jmarshall: Argh… I didn't think to strip quote marks in the latest edition of that workflow regex…

PR #233, opened 2025-06-19
Add missing version labels to images

MattWellie:
Still missing a suitable label for STR-r

Deletes Talos which is no longer built from this repository


jmarshall:
That seems like a scenario that @nevoodoo's script ought to be able to handle without disrupting any other images in the PR…

nevoodoo:
I think we might want to remove the print statement for when the file is not found in case that interferes with picking up the json, #234 might be a better fix for this

jmarshall:
If you rebase this PR onto **main** now, you should not have the failure due to missing Talos.

PR #234, opened 2025-06-19
Ignore Dockerfiles that have been deleted by the pull request

jmarshall:
See PR #233, in which one deleted _Dockerfile_ has stopped the workflow from working for other updated images.


nevoodoo:
Looks good! Thanks for the fix and apologies for the oversight - don't think I accounted for deleted images at the time.

nevoodoo (approved):

PR #235, opened 2025-06-20
Add Manta dockerfile that disables RRR for INS detection

EddieLF:
See https://github.com/populationgenomics/gatk-sv/pull/30


MattWellie (approved):
🙏

PR #241, opened 2025-07-11
Update bcftools_121 to install htslib

EddieLF:
Install htslib so that the `bgzip` command can be used. Without installing htslib, this command is not found. E.g. [this job](https://batch.hail.populationgenomics.org.au/batches/626478/jobs/1).


jmarshall:
You can use `… | bcftools sort -Oz -o …/vcf_out.vcf.gz` in that job to avoid needing to use the separate `bgzip` executable. (And that filename extension probably ought to be `.vcf.gz` rather than `.vcf.bgz` unless you're interoperating with tools with very unfortunate inflexible naming conventions.)

Building bcftools also builds htslib. So if you still want to install HTSlib's tools in this image too, instead of separately downloading and building another htslib, you can just add another `make install` line after the existing one:

```
make -C htslib-$VERSION DESTDIR=/bcftools_install install
```

Finally, it's time to move away from hackily versioned images like `bcftools_121`. There should be no reason not to do this in _images/bcftools/Dockerfile_ instead and delete _images/bcftools_121/**_.

MattWellie:
I've posted this for discussion on Slack [here](https://centrepopgen.slack.com/archives/C018KFBCR1C/p1752214099479879)

Agreed that we want one BCFtools to rule them all (binning off _120 and _121), but we still have to address the original reason these names were created - directly changing the `bcftools` Dockerfile alters the default image across all workflows which use `image_path`, and we need agreement across users for that.

PR #242, opened 2025-07-11
Update bcftools to 1.22, remove bcftools_121/Dockerfile

EddieLF:
- Updates bcftools to latest version (1.22)
- Refactors the bcftools Dockerfile with the build process originally implemented in https://github.com/populationgenomics/images/pull/177. 
- Removes `/images/bctools_121/Dockerfile` as there's no longer any need to maintain multiple Dockerfiles and images for different versions of the same tool.


EddieLF:
I like that idea @MattWellie and I've updated this PR to reflect that, start with 1.20 and then two small follow up PRs to build and deploy the other versions to the artifact repo where they will live for as long as we need them.

MattWellie (approved):
LGTM!

PR #245, opened 2025-07-22
Update Dockerfile - add lme4

htanudisastro:


images/r-meta/Dockerfile
```diff
@@ -20,9 +20,11 @@ RUN apt update && \
         r-argparser \
         r-arrow \
         r-base \
+	r-broom.mixed \
```
jmarshall, line 23: The added lines are made with tabs instead of spaces, but no biggie 🤷

PR #248, opened 2025-07-29
Update Dockerfile

annacuomo:
Update dockerfile to contain latest change to SAIGE-QTL

see https://centrepopgen.slack.com/archives/C030X7WGFCL/p1753783940791109?thread_ts=1726615809.665199&cid=C030X7WGFCL


images/saige-qtl/Dockerfile
```diff
@@ -3,6 +3,8 @@ FROM australia-southeast1-docker.pkg.dev/analysis-runner/images/driver-r:1.4
 # Override this with a specific SHA in images.toml, as the qtl fork doesn't have releases.
 ARG VERSION=HEAD
```
MattWellie, line 4: Probably delete these two lines, you don't want ambiguous values for VERSION

MattWellie (approved):

jmarshall (approved):
I've verified that everything upstream's _pixi.toml_ wants to install ends up installed (via R dependencies listed in _DESCRIPTION_) and the installed scripts have been edited to not use `pixi` and do run (at least enough to not fail with `library()` or `require()` errors!).

PR #252, opened 2025-09-01
Install procps in cpg_hail and multiqc

EddieLF:
Install `procps` in the cpg_hail image

- images using `cpg_hail` will need to update the version to match this one if they want procps to be included


MattWellie (approved):
LGTMatt

PR #292, opened 2026-03-30
Using base image with Python for bcftools

michael-harper:
Popgen is refactoring their codebase and requires bcftools to have access to python so that jobs can call scripts that use bcftools via python `subprocess`


Alexander-Stuckey (changes requested):
version in `main` is 1.23

Alexander-Stuckey (approved):
Looks good to me

PR #295, opened 2026-04-23
VNtyper image from cpg_hail_gcloud:0.2.137.cpg1-2

EddieLF:
New image that installs [VNtyper](https://github.com/hassansaei/VNtyper) v2.0.3

- Use Micromamba to install dependencies, namely `bwa v0.7.18` and `samtools v1.20`, ripped from [the Conda requirements](https://github.com/hassansaei/VNtyper/blob/main/conda/environment_vntyper.yml)
- Clone the repo install it by following [the README instructions](https://github.com/hassansaei/VNtyper/tree/main#using-pip)
- Install hg38 references, no need for hg19 for our workflows
- Make a minor update to the source code, which sets the CRAM reference to depend on an environment variable that we can inject via the pipeline. Without this, [the CRAM reference is hardcoded to be an empty string](https://github.com/hassansaei/VNtyper/blob/main/vntyper/scripts/fastq_bam_processing.py#L124). By letting it depend on an env variable, we can ensure the right reference fasta is used like this:
```python
    reference = hail_batch.fasta_res_group(batch_instance)

    job.command(f"""\
    export CRAM_REFERENCE={reference['base']}
    vntyper ...
    """)
```
- Create symlink for the vntyper dir, otherwise it doesn't know where to find the kestrel jar (see https://batch.hail.populationgenomics.org.au/batches/1128850/jobs/1)

### DIFF
--- a/images/vntyper/Dockerfile
+++ b/images/vntyper/Dockerfile
@@ -22,15 +22,13 @@ RUN git clone https://github.com/hassansaei/vntyper.git && \
     git checkout v${VERSION} && \
     pip install --no-cache-dir --upgrade pip && \
     pip install --no-cache-dir . && \
-    mkdir -p /vntyper/references && \
-    vntyper install-references --output-dir /vntyper/references --references hg38
+    vntyper install-references --output-dir /vntyper/reference --references hg38
 
 # Patch vntyper to use the CRAM_REFERENCE environment variable, necessary for compatibility with CPG's CRAM reference management
 RUN sed -i 's|cram_ref_option = ""|cram_ref_option = f"--reference {os.environ['"'"'CRAM_REFERENCE'"'"']}" if os.environ.get("CRAM_REFERENCE") else ""|' \
       /usr/local/lib/python3.11/site-packages/vntyper/scripts/fastq_bam_processing.py
 
-# Create symlinks for vntyper and its references in the vntyper package install directory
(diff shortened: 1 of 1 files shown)


images/vntyper/Dockerfile
```diff
@@ -0,0 +1,36 @@
...
+    mkdir -p /vntyper/references && \
+    vntyper install-references --output-dir /vntyper/references --references hg38
+
+# Patch vntyper to use the CRAM_REFERENCE environment variable, necessary for compatibility with CPG's CRAM reference management
+RUN sed -i 's|cram_ref_option = ""|cram_ref_option = f"--reference {os.environ['"'"'CRAM_REFERENCE'"'"']}" if os.environ.get("CRAM_REFERENCE") else ""|' \
+      /usr/local/lib/python3.11/site-packages/vntyper/scripts/fastq_bam_processing.py
+
+# Create symlinks for vntyper and its references in the vntyper package install directory
+RUN ln -s /usr/local/lib/python3.11/site-packages/vntyper /vntyper/vntyper && \
+    ln -s /vntyper/references /vntyper/reference
```
MattWellie: this is interesting - the output directory for the references is set on line 26 in this file. Can't we just write the references out to `/vntyper/reference` instead of `/vntyper/references` and remove the need for this symlink?

EddieLF: Yup good shout

MattWellie (approved):
Rapid work

PR #296, opened 2026-04-28
Vntyper

EddieLF:
Update the VNtyper image to install advntr

- Install `gcc` and `g++` for the advntr install
- Use Micromamba 2.5.0 instead of ancient 0.8.2
- Also install bcftools to suppress warnings and allow compressed vcf outputs
- Download the advntr db references during the `install-references` command to allow for using code-advntr tool
- Install adVNTR and its dependencies into a separate conda env called `envadvntr` - this tool uses python 2 and the vntyper pipeline requires the separate environment
  - Borrow some code from the vntyper `install_advntr.sh` shell script to handle how newer GCC versions compile C code - the shell script doesn't work out of the box due to requiring `conda activate` commands, so install the env using micromamba directly


MattWellie (approved):
Image

PR #297, opened 2026-04-30
Update cpg_hail image to use 0.2.138.cpg1 hail tag

EddieLF:
Following release of the latest Hail upgrade: https://github.com/populationgenomics/hail/tags

### DIFF
--- a/images/cpg_hail/Dockerfile
+++ b/images/cpg_hail/Dockerfile
@@ -13,7 +13,7 @@ RUN apt-get update && \
         git \
         gnupg \
         jq \
-        openjdk-11-jdk-headless \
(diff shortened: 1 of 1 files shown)


MattWellie (approved):

images/cpg_hail/Dockerfile
```diff
@@ -13,7 +13,7 @@ RUN apt-get update && \
         git \
         gnupg \
         jq \
-        openjdk-11-jdk-headless \
+        openjdk-17-jre-headless \
```
jmarshall, line 16: You want to have **both** JDK 11 and JRE 17.

PR #300, opened 2026-05-12
Add KING 2.3.2 image

joshuamschmidt:
## Summary
- Adds `images/king/Dockerfile` for [KING](https://www.kingrelatedness.com/) 2.3.2 (latest stable, released 2023-09-08), needed for upcoming `popgen-genotyping` `--ibdseg` relatedness inference work (IBD0/IBD1/IBD2 estimates).
- version control by getting pre-compiled binaries from https://www.kingrelatedness.com/history.shtml.
- The Dockerfile derives the upstream URL token from `ARG VERSION` by stripping dots (`2.3.2` → `Linux-king232.tar.gz`), so future version bumps only need a one-line `ARG VERSION=...` change.
- KING's binary dynamically links against `libquadmath0` and `libgomp1`, neither of which ship in `python:3.11-slim-bookworm` — both are added via apt.

## Test plan
- [x] Local `docker build --platform linux/amd64 -t king:test images/king/` succeeds.
- [x] `docker run --rm --platform linux/amd64 king:test king` prints `KING 2.3.2` banner and the expected "Genotype files are required" error.
- [x] `ldd /usr/local/bin/king` inside the built image resolves all shared libraries.
- [x] CI deploy workflow builds and pushes `australia-southeast1-docker.pkg.dev/cpg-common/images/king:2.3.2-1` on merge.


images.toml
```diff
@@ -27,6 +27,7 @@ hap-py = '0.3.15'
 haplocheckcli = '64293b3481d52025a01a293f6cc891546c30660e'
 ica = '2.31.8'
 jasmine = '1.1.5'
+king = '2.3.2'
```
MattWellie, line 30: No biggie, but this images toml is no longer used to generate the standard/global config. I don't think it's used for anything any more.

Fine to add to it, but the resulting king image won't be available in default analysis-runner configs

MattWellie (approved):
LGTMatt

images.toml
joshuamschmidt: oh - I haven't created a new image in a while...... 

> resulting king image won't be available in default analysis-runner configs  

What should I do instead?

PR #304, opened 2026-05-27
Latest version & HTSlib fix

MattWellie:
See https://github.com/samtools/htslib/blob/develop/NEWS#L4


Alexander-Stuckey (approved):
Looks good to me

PR #306, opened 2026-05-29
nf-base bcftools 1.23.1

MattWellie:
BCFtools install based on the nf-generic candidate. Should be able to deploy as a NF task, be monitored with procps, and log results with metamist.

It's the same as our current bcftools image, but on a different base, and with the compile libraries already installed in nf-base stripped out

### DIFF
--- a/images/nf_bcftools/Dockerfile
+++ b/images/nf_bcftools/Dockerfile
@@ -1,0 +1,45 @@
+FROM australia-southeast1-docker.pkg.dev/cpg-common/images-dev/nf_base:0.1.0-1 AS base
+
+ARG VERSION=1.23.1
+
+LABEL \
+  version="v${VERSION}" \
+  description="NF Base Bcftools install"
+
+RUN apt-get update && apt-get install -y \
+        libbz2-1.0 \
+        libcurl4 \
+        liblzma5 \
+        libssl3 \
+        zlib1g && \
+    rm -r /var/lib/apt/lists/* && \
+    rm -r /var/cache/apt/*
+
+FROM base AS compiler
+
+RUN apt-get update && apt-get install -y \
+        libbz2-dev \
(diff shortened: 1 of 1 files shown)


images/nf_bcftools/Dockerfile
```diff
@@ -0,0 +1,45 @@
...
+    cd bcftools-${VERSION} && \
+    # Remove any existing array to vcf plugins and replace with the latest from the gtc2vcf repository
+    /bin/rm -f plugins/{idat2gtc.c,gtc2vcf.{c,h},affy2vcf.c} && \
+    wget -P plugins \
+    https://raw.githubusercontent.com/freeseek/gtc2vcf/master/idat2gtc.c \
+    https://raw.githubusercontent.com/freeseek/gtc2vcf/master/gtc2vcf.c \
+    https://raw.githubusercontent.com/freeseek/gtc2vcf/master/gtc2vcf.h \
+    https://raw.githubusercontent.com/freeseek/gtc2vcf/master/affy2vcf.c \
+    https://raw.githubusercontent.com/freeseek/gtc2vcf/master/BAFregress.c && \
+    ./configure --enable-libcurl --enable-s3 --enable-gcs && \
```
Alexander-Stuckey, line 37: we _probably_ don't need to enable s3 support in our bcftools image, but I guess it doesn't hurt

Alexander-Stuckey (approved):
Looks good to me, dunno if you care about s3 support

images/nf_bcftools/Dockerfile
MattWellie: yeah I was wondering about that... all the gtc2vcf content is for the popgen genotyping arrays, which is not universally applicable, but in total this whole dockerfile install adds 35MB to the nf-base so happy to keep it all in

PR #312, opened 2026-07-06
Update to VerifyBamID, and much smaller overall image

MattWellie:
(not sure if PopGen use this image)

n.b. this includes a fix that might be worth upstreaming - the cmake & test loop built into the upstream dockerfile was broken on my local machine, and after a lot of back and forth it looks like a platform-specific failure mode. From Claude:

```
- On x86_64, char defaults to signed, so the code's -1 sentinel (missing genotype) already worked correctly there — that's likely why this Dockerfile worked before.          
  - On aarch64 (e.g., your Mac's Docker), char defaults to unsigned per the ARM ABI, which silently turned the -1 sentinel into 255, breaking missing-genotype detection.       
  - -fsigned-char is a standard GCC/Clang flag supported on all architectures (x86_64, aarch64, etc.) that explicitly forces char to be signed, overriding whatever the platform
  default is.                                                                                                                                                                   
                                                                                                                                                                                
  So adding this flag makes the build behave identically on both amd64 and arm64 — it's not a workaround specific to your machine, it fixes a genuine latent portability bug in 
  the upstream code that only happened to be masked on x86_64.
```


I'm ambivalent about upstreaming, nobody is productionising workflows on mac, and this repo is barely maintained, but this change brings the docker image size down 90%, whilst updating the version and upgrading HTSlib from 1.11 -> 1.23.1

### DIFF
--- a/images/verifybamid/Dockerfile
+++ b/images/verifybamid/Dockerfile
@@ -1,14 +1,55 @@
-FROM debian:buster-slim
+# In this image we do not include the R-based plot scripts.
+FROM ubuntu:20.04 AS build
 
-ENV MAMBA_ROOT_PREFIX /root/micromamba
-ENV PATH $MAMBA_ROOT_PREFIX/bin:$PATH
-ARG VERSION=${VERSION:-2.0.1}
+# Set noninterative mode
+ENV DEBIAN_FRONTEND=noninteractive
+ENV LD_LIBRARY_PATH="/usr/local/lib"
 
-RUN apt-get update && apt-get install -y git wget bash bzip2 zip && \
-    rm -r /var/lib/apt/lists/* && \
-    rm -r /var/cache/apt/* && \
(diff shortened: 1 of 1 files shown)


images/verifybamid/Dockerfile
```diff
@@ -1,14 +1,55 @@
...
-    rm -r /root/micromamba/pkgs
+# In this image we do not include the R-based plot scripts.
+FROM ubuntu:20.04 AS build
+
+# Set noninterative mode
+ENV DEBIAN_FRONTEND=noninteractive
+ENV LD_LIBRARY_PATH="/usr/local/lib"
+
+ARG HTSLIB_VERSION="1.23.1"
+ARG VERSION="v2.0.3"
```
MattWellie:
```suggestion
ARG HTSLIB_VERSION=1.23.1
ARG VERSION=v2.0.3
```

EddieLF (approved):

PR #318, opened 2026-08-13
Update image path 

htanudisastro:
I think the terra account does not have access to `cpg-common` based on the error message in: https://centrepopgen.slack.com/archives/C030X7WGFCL/p1786580947459439. 

Would like to rebuild in `australia-southeast1-docker.pkg.dev/tenk10k-sv/images/sv` (directory where all the other images that Terra pulls from are stored in)


Alexander-Stuckey (approved):
Looks good to me

PR #324, opened 2026-08-18
Bump plink2 to alpha7 build 20260818 (chrX --hwe AVX2 fix)

joshuamschmidt:
## What

Updates the plink image to use plink2 alpha 7.4, build 20260818 (was 20260808).

## Why

The AVX2 builds of plink2 had a bug in missing-sex handling. This made chrX `--hwe` hang forever (or crash in older builds) on some datasets. See [plink-ng issue #337](https://github.com/chrchang/plink-ng/issues/337).

The 18 Aug 2026 build fixes this. From the plink2 changelog:

> Fixed a6/a7 AVX2 missing-sex handling bug which potentially affected chrX --hardy/--hwe, chrY --glm, chrY --geno-counts, and --y-nosex-missing-stats.

## Changes

- All three plink2 downloads (Intel AVX2, AMD AVX2, generic x86_64) bumped to build 20260818
- Image version tag updated to `1.9-20250819-PLINK-2.0-20260818`
- PLINK 1.9 unchanged

### DIFF
--- a/images/plink/Dockerfile
+++ b/images/plink/Dockerfile
@@ -2,7 +2,11 @@
 FROM python:3.11-slim-bookworm
 
 # Build metadata
+# CI's get_version.py derives the image tag by text-matching the first
+# VERSION assignment in this file (comments included), so keep the value
+# below literally spelt out — no variables.
 ENV VERSION=1.9-20250819-PLINK-2.0-20260818
+ARG PLINK2BUILD=20260818
 LABEL maintainer="Population Genomics Team"
 LABEL version="${VERSION}"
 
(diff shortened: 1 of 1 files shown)


michael-harper (approved):
LGTM

images/plink/Dockerfile
```diff
@@ -2,7 +2,7 @@
 FROM python:3.11-slim-bookworm
 
 # Build metadata
-ENV VERSION=1.9-20250819-PLINK-2.0-20260808
+ENV VERSION=1.9-20250819-PLINK-2.0-20260818
```
jmarshall, line 8: This might be a good time to change this to

```suggestion
ENV VERSION=1.9-20250819-PLINK-2.0-20260818
ARG PLINK2BUILD=20260818
```

and use `$PLINK2BUILD` in the wget lines below.

Except that people updating it need to remember to leave the `ENV VERSION=…` line literally spelt out (for the sake of the _get_version.py_ script), so this might be more trouble than it's worth.

jmarshall (approved):
You could remove the `PLINK2_FORCE_VARIANT` workaround from the script so callers will get the good stuff with this update… but I guess it would be wiser to leave it in in case of future problems!

joshuamschmidt:
> You could remove the `PLINK2_FORCE_VARIANT` workaround from the script so callers will get the good stuff with this update… but I guess it would be wiser to leave it in in case of future problems!

Agreed on leaving it in. The default is still auto-selection, so this update gets callers the fixed AVX2 build regardless. The explicit generic pin (`job.env('PLINK2_FORCE_VARIANT', 'generic')`) I used only in `popgen-genotyping`'s HWE job and I'll drop it there once this image update merges.

images/plink/Dockerfile
joshuamschmidt: Done in a9aed23. Added `ARG PLINK2BUILD` as you suggested. RE `ENV VERSION` literal, left a comment warning future editors why. Fun footnote: get_version.py's regex doesn't skip comments, so my first attempt at the warning comment (which quoted "ENV VERSION=…") was itself picked up as the version. Reworded to avoid the token sequence.

jmarshall: Nice one.

We should probably fix the script to strip comments — it's the sort of thing you only find out the hard way…

